PR #7, opened 2015-03-12
change DELAY and WIDTH of pulse_gen from 16bits to 32bits

thirono:
The bit widths of DELAY and WIDTH of pulse_gen were changed from 16bits to 32bits.


themperek:
Next time would be nice if you also add some test to check what was done :-)

PR #13, opened 2015-03-20
BUG: fix possible bug in length calculation

laborleben:
ENH: repeat and wait increased to 32bit


themperek:
Could you update documentation/Readme too.

laborleben:
updated the documentation.

is the change
reg [7:0] status_regs [8+MEM_BYTES*2-1:0];
to 
reg [7:0] status_regs [16-1:0];
breaking anything?

themperek:
Should not. Need to check.

themperek:
Now it is something else. We use spi for StdRegister and now  it does not match anymore.

laborleben:
the module version check fails during init (read:0, require: ==1). Any idea?

themperek:
Make some hack to disable this!

Can remove/change this requirement. Can do some hack like (before init)?
```python
dut['spi']._require_version = ">=0"
dut.init() 
```
Has to also change address now it from 16 not 8 like before.

laborleben:
fixed reading version by adding mem init parameter to Dummy

PR #66, opened 2016-11-29
Added data throughtput test to the MMC3 ethernet example

MarcoVogt:
Files to be merged:
examples/mmc3_eth/mmc3_eth_throughput_test.py
examples/mmc3_eth/mmc3_eth.yaml 
examples/mmc3_eth/firmware/src/mmc3_eth_throughput_test.v

The other files are wip (spi <--> i2c eeprom)


themperek:
Cool. Thank you.

I would just suggest small changes if you can (just change you branch they will update automatically):
-  make mmc3_eth_throughput_test.v just mmc3_eth.v do not think we need both. Same for python and yaml files just keep one version (with all features).
- spi_emu_simple.v i would name spi_eeptrom_emu.v or similar and put into firmware/modules/utils directory. Make utils.v inside this single verilog file. To many files/includes may be confusing.

Other opinions/points of view welcome.

themperek:
You can add small README.rst into https://github.com/SiLab-Bonn/basil/tree/master/examples/mmc3_eth with throughput result and few lines explanation see [mmc3_m26_eth ](https://github.com/SiLab-Bonn/pyBAR/tree/master/firmware/mmc3_m26_eth).

MarcoVogt:
Actually, I'd like to pull, but it seems I have no write permissions. Can you give me write access to the dev branch?
The SPI/I²C stuff is wip, so better pull it later.

themperek:
>The SPI/I²C stuff is wip, so better pull it later.

spi_emu_simple.v/utils.v is still there ?

Can you remove: test_mmc3_eth.yaml ?

Can you add to readmy: "The Firmware make use of free [SiTcp](http://sitcp.bbtech.co.jp/) Ethernet module." ?

MarcoVogt:
The utils belong to the spi stuff and who knows if this will be used anyhow ;). Why? Because SiTCP offers a similar thing for an M24C08 chip on their website as I figured out after writing it. At some point I will evaluate theirs and if it's good, why not: https://sitcp.bbtech.co.jp/modules/d3downloads/index.php?page=singlefile&cid=1&lid=18

2 and 3 I will edit now.

themperek:
Can you remove then spi_emu_simple.v/utils.v files.

M24C08 from SiTCP looks like something that may work for us. Would be good to know what is the license for those files.

themperek:
Thank you.

PR #69, opened 2017-01-10
Update SiTcp.py

MarcoVogt:
Increased buffer size of _tcp_readout to improve data throughput.


laborleben:
Can you add upstream branch so that we can checkout?

PR #72, opened 2017-07-28
Si570 programmable oscillator

MarcoVogt:
Basil driver, firmware and tests written by Matthieu Batista (CPPM).


MarcoVogt:
Needs some rework, I guess

Issue #78, opened 2018-01-12
Module level logging


DavidLP:
Effective logging needs tunable filtering. Using the root logger prevents tunable filtering. Thus I propose to use [module level logging](https://docs.python.org/2/howto/logging.html).

DavidLP:
Implemented in development

PR #79, opened 2018-01-12
ENH: use module level logging

DavidLP:
Addresses issue #78.

We used just the root logger before for every module in basil.

Detailed example:

When using a laboratory SCPI device one would get such a message:

```bash
2018-01-12 10:05:40,877 - root - [INFO    ] (MainThread) BASIL VISA TL with @sim backend found the following devices
```
`root` stands for the root logger. If one is not interested in messages of this SCPI devices, but one wants to concentrate on the DUT there is no way thus far to change/suppress this info, when using the root logger in the DAQ application. With this PR a module based logger is implemented that allows to fine tune the log per module. The message reads then:

```bash
2018-01-12 10:11:48,293 - basil.TL.Visa - [INFO    ] (MainThread) BASIL VISA TL with @sim backend found the following devices
```

Now it is possible in your application to define the loglevel of this module logger to suppress this info with

```python
import logging
logging.getLogger('basil.TL.Visa').setLevel(logging.WARNING)
```

### DIFF
--- a/basil/HL/RegisterHardwareLayer.py
+++ b/basil/HL/RegisterHardwareLayer.py
@@ -62,7 +62,7 @@ class RegisterHardwareLayer(HardwareLayer):
             version = str(self.VERSION)
         else:
             version = None
-        logger.debug("Initializing %s (firmware version: %s), module %s, base_addr %s" % (self.name, version if 'VERSION' in self._registers else 'n/a', self.__class__.__module__, hex(self._base_addr)))
+        logger.info("Initializing %s (firmware version: %s), module %s, base_addr %s" % (self.name, version if 'VERSION' in self._registers else 'n/a', self.__class__.__module__, hex(self._base_addr)))
         if self._require_version and not eval(version + self._require_version):
             raise Exception("FPGA module %s does not satisfy version requirements (read: %s, require: %s)" % (self.__class__.__module__, version, self._require_version.strip()))
         for reg, value in self._registers.iteritems():
--- a/basil/HL/RegisterHardwareLayer.py
+++ b/basil/HL/RegisterHardwareLayer.py
@@ -14,6 +14,8 @@ from collections import namedtuple
 from basil.utils.BitLogic import BitLogic
 from basil.HL.HardwareLayer import HardwareLayer
 
+logger = logging.getLogger(__name__)
+
 
 # description attributes
 read_only = ['read_only', 'read-only', 'readonly', 'ro']
@@ -60,7 +62,7 @@ class RegisterHardwareLayer(HardwareLayer):
             version = str(self.VERSION)
         else:
             version = None
-        logging.debug("Initializing %s (firmware version: %s), module %s, base_addr %s" % (self.name, version if 'VERSION' in self._registers else 'n/a', self.__class__.__module__, hex(self._base_addr)))
+        logger.debug("Initializing %s (firmware version: %s), module %s, base_addr %s" % (self.name, version if 'VERSION' in self._registers else 'n/a', self.__class__.__module__, hex(self._base_addr)))
         if self._require_version and not eval(version + self._require_version):
             raise Exception("FPGA module %s does not satisfy version requirements (read: %s, require: %s)" % (self.__class__.__module__, version, self._require_version.strip()))
         for reg, value in self._registers.iteritems():
@@ -188,14 +190,14 @@ class RegisterHardwareLayer(HardwareLayer):
             try:
                 return self._get(attribute)
             except Exception, e:
-                logging.error(e)
+                logger.error(e)
                 return None
 
         def setter(self, value):
             try:
                 return self._set(attribute, value)
             except Exception, e:
-                logging.error(e)
+                logger.error(e)
                 return None
         # construct property attribute and add it to the class
         setattr(self.__class__, attribute, property(fget=getter, fset=setter, doc=attribute + ' register'))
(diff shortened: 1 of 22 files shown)


basil/HL/RegisterHardwareLayer.py
```diff
@@ -60,7 +62,7 @@ def init(self):
             version = str(self.VERSION)
         else:
             version = None
-        logging.debug("Initializing %s (firmware version: %s), module %s, base_addr %s" % (self.name, version if 'VERSION' in self._registers else 'n/a', self.__class__.__module__, hex(self._base_addr)))
```
themperek: Can you change this to info back?

basil/utils/sim/Test.py
```diff
@@ -37,7 +39,7 @@ def socket_test(dut, debug=False):
     port = os.getenv("SIMULATION_PORT", '12345')
 
     if debug:
-        dut._log.setLevel(logging.DEBUG)
```
themperek: This is this cocotb thing so I am not sure. Better not to touch.

basil/dut.py
```diff
@@ -14,6 +12,12 @@
 import warnings
 from collections import OrderedDict
 
+# FIXME: Bad practice
+# Logger settings should not be defined in a module, but once by the
+# application developer. Thus outside of basil. Otherwise multiple calls to
+# the basic config are possible. This is left here at the moment for backward
+# compatibility and since our logging format is the same everywhere (?).
+import logging
```
laborleben: I find this comment ridiculous. Just saying something is bad practice that served us well for many years. Keep in mind that basil is still a hardware debugging and development framework.
Maybe your application and not the framework is doing something wrong. You still can import basil into your application and change the behavior of the global logger. Something that you should consider than changing the behavior for everyone else.

laborleben:
Now we have a logger per module. What has changed? Can you explain?

basil/dut.py
DavidLP: This comment gives a reason why this should be removed in future and thus is a FIXME. What is the issue here?

themperek:
For me, this looks better than before.

basil/dut.py
DavidLP: I give you an example why this is bad practice:

```python

from basil.dut import Dut

def my_func():
    logging.info('Bla')

if __name__ == '__main__':
    import logging
    logging.basicConfig(level=logging.INFO, format="My format %(message)s")
    my_func()
```

The own defined format is overwritten on basil.dut import. No think about several projects, each defining a logging format as basil does. The module import order would define the logging format.

basil/utils/sim/Test.py
DavidLP: Yes, was not sure about this either

laborleben:
The formatting is not sufficient. Every logger output has a different indention when coming from different (basil-)modules because the logger name is different (and can be very long). This is an improvement for debugging but a disadvantage for readability (think of the limited space in the terminal).
Limiting the space for the logger name (so far it was only "root") and printing only the last part of the module name could solve these issues.

laborleben:
I still appreciate the other way around: Fix your application (you can set up as many loggers as you want, give the pet names...) and changing the behavior of the root logger by setting up the log level just once.

DavidLP:
> I still appreciate the other way around: Fix your application (you can set up as many loggers as you want, give the pet names...) and changing the behavior of the root logger by setting up the log level just once.

Right now it is not possible to suppress logging outputs per module individually. That is a very reasonable feature for a framework consisting of several modules. This cannot be done in the application using basil.

> The formatting is not sufficient. Every logger output has a different indention when coming from different (basil-)modules because the logger name is different (and can be very long). This is an improvement for debugging but a disadvantage for readability (think of the limited space in the terminal).
Limiting the space for the logger name (so far it was only "root") and printing only the last part of the module name could solve these issues.

It is the simplest way of doing it and I personally prefer simplicity by just using the module  `__name__` attribute. This is a standard and can for example be found in the official [logging manual](https://docs.python.org/2/howto/logging.html#advanced-logging-tutorial). Right now this also does not increase the log line width too much. I agree, if this becomes an issue we can redefine the logging naming scheme.

basil/dut.py
DavidLP: Just deleting the logging import is also a solution, but applications using basil development would suddenly not see any logs or logs formatted different (depending on their logging configuration). We could say it is development branch so we expect other project to comply. But this step I did not want to do here. But if you think that is better I'll just delete the lines.

laborleben:
```PYTHON
import logging

# just set up the basil logger format and level here
logging.basicConfig(level=logging.ERROR)
from basil.dut import Dut

# Your own logger here
logger = logging.getLogger("test")
logger.propagate = False
logger.setLevel(logging.INFO)
console = logging.StreamHandler()
console.setLevel(logging.INFO)
formatter = logging.Formatter("My format %(message)s")
console.setFormatter(formatter)
logger.addHandler(console)

dut = Dut("")
dut.init()

def my_func():
    logging.error("logging info message")
    logger.info('logger info message')
    logger.error('logger error message')

if __name__ == '__main__':
    my_func()
```

laborleben: So you still want to try to fix the output in you application by changing the underlying framework. Good, but the above code snippet can be used in you application without changing basil.

DavidLP: How do you set the logging of a basil module with your approach?

PR #80, opened 2018-04-09
ENH: increase bits of trigger data delay

YannickDieter:
This will increase the number of bits for `TRIGGER_DATA_DELAY` from 4 to 8 bit in order to be able to set the correct trigger data delay for faster clocks.


YannickDieter:
Has anybody an idea why travis fails? Is it possible to restart the build? It looks like it is due to some travis problems and not due to my added code.

YannickDieter:
Thank you!

PR #82, opened 2018-04-24
SiTCP improvements

laborleben:
- Reducing issues with UDP socket
- Retry send and recv on failure
- Set sockets to blocking (use select instead)
- Set UDP recv buffsize to expected value
- Use message ID for UDP messages
- Adding tcp_to_bus.v module (support both, UDP writing/reading and TCP writing to/from BUS)
- Adding tcp_to_bus support to SiTcp transfer layer
- Increase write speed by 50% when using tcp_to_bus
- Update ```test_eth.py``` script

This update is addressing the following issues:
- Use select to check whether the socket is ready to send or receive data (this was not the case before)
- The UDP message was successfully sent but not received (return message missing) -> resend message
- The UDP message was successfully sent but received wrong return message (message is a clone from one of the previous return messages, message was sent twice by SiTCP) -> read one more time
- The UDP interface becomes busy for several seconds -> retry sending message
- Check for pending messages before send and after recv, check message ID, and give warning message

The update is NOT addressing the following issues:
- Message and return message are OK, but reading back register gives wrong value -> **rare**
- The message was successfully sent but the return message never arrived -> resend message
(this is especially critically in case of a "start FSM" register) -> **very rare**
- Duplicate, missing and out of order packages are generally not handled correctly because message ID is not available in FPGA.

Solution:
Use tcp_to_bus.
```
transfer_layer:
  - name  : ETH
    type  : SiTcp
    init:
        ip : "192.168.10.16"
        udp_port : 4660
        tcp_port : 24
        tcp_connection : True
        tcp_to_bus : True
```

### DIFF
--- a/basil/HL/sitcp_fifo.py
+++ b/basil/HL/sitcp_fifo.py
@@ -4,6 +4,9 @@
 # SiLab, Institute of Physics, University of Bonn
 # ------------------------------------------------------------
 #
+import struct
+import array
+
 import numpy as np
 
 from basil.HL.HardwareLayer import HardwareLayer
@@ -56,7 +59,7 @@ class sitcp_fifo(HardwareLayer):
             super(sitcp_fifo, self).__setattr__(name, value)
 
     def get_data(self):
-        ''' Reading data from SiTCP FIFO.
+        ''' Reading data from SiTCP FIFO (via TCP).
 
         Returns
         -------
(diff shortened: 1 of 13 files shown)


themperek:
Can you use logger instead for logging?

themperek:
Would it be possible to integrate the test_eth into: https://github.com/SiLab-Bonn/basil/tree/development/examples/mmc3_eth (because this is under CI)

laborleben:
Done and works!

PR #87, opened 2018-06-29
ENH: allow sharing of fast trigger with other tdc modules

YannickDieter:
This allows to share the fast sampled trigger signal (in case of `FAST_TRIGGER=1`) with other TDC modules by setting the `BROADCAST` parameter.


YannickDieter:
Yes sure :)

YannickDieter:
Do we want to have a unit test for the broadcasting, I think it would be nice.
I can test if all tdc module measures the same result when using one common trigger and the same TDc signal.

YannickDieter:
Can we merge?

laborleben:
What is the outcome of the test that you proposed?

YannickDieter:
Ah.. yes. I see, this was the missing part, I forgot. I will add a test.

YannickDieter:
I pushed a test for tdc broadcasting. If you are fine with that you can merge.

YannickDieter:
I pushed a test for tdc broadcasting. If you are fine with that you can merge.

YannickDieter:
I added a test for tdc broadcasting. If you are fine with this you can merge.

PR #88, opened 2018-07-03
ADD_GPIB

Marrkson:
ADD: This pull adds the support for the Prologix USB to GPIB controller as described here:
http://prologix.biz/downloads/PrologixGpibUsbManual-4.2.pdf
This enables GPIB usage with basil for linux

### DIFF
--- a/basil/HL/hp_e3632a.yaml
+++ b/basil/HL/hp_e3632a.yaml
@@ -1,0 +1,20 @@
+# Device description for HP E3632A Powersupply.
+# set_ function expect a parameter, get_ function return a parameter.
+# Just the very basic commands are imlemented here.
+identifier : HEWLETT-PACKARD,E3632A,0,1.1-5.0-1.0
+on : OUTP ON
+off : OUTP OFF
+disp_on : DISP 1
+disp_off : DISP 0
+get_current : MEAS:CURR?
+set_current : SOUR:CURR
+set_voltage : SOUR:VOLT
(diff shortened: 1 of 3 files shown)


themperek:
@DavidLP Can you look at this?

basil/TL/PrologixVisa.py
```diff
@@ -0,0 +1,53 @@
+#
```
DavidLP: As I see it the only difference to the already existing VISA implementation is the query command. So one could simplify this module to:
``` python

class PrologixVisa(Visa):
    def query(self, data):
        self._resource.write("++auto 0")
        self._resource.write(data)
        return self._resource.query("++read eoi")
```

and it would do the same, no?
I am not sure if the _write_ and _read_ commands also need changes. Did you check?

basil/HL/hp_e3632a.yaml
```diff
@@ -0,0 +1,20 @@
+# Device description for HP E3632A Powersupply.
+# set_ function expect a parameter, get_ function return a parameter.
+# Just the very basic commands are imlemented here.
+identifier : HEWLETT-PACKARD,E3632A,0,1.1-5.0-1.0
```
DavidLP: The identifier does not have to be fully matched. So sometimes it is better to not specify minor firmware versions for compatibility. I guess _HEWLETT-PACKARD,E3632A_ could be enough here?

PR #95, opened 2019-01-07
ENH: Move firmware folder to basil (part of pakage)

themperek:


setup.py
```diff
@@ -21,6 +22,13 @@
 with open('requirements.txt') as f:
     install_requires = f.read().splitlines()
 
+def package_files(directory):
```
DavidLP: Educated "3rd" best Stackoverflow answer, that acutally is the best answer ;-)

themperek: ?

themperek:
Package size should not change since firmware was already before part of it just not installed.

setup.py
DavidLP: https://stackoverflow.com/questions/27664504/how-to-add-package-data-recursively-in-python-setup-py/27664856

laborleben: Isn't MANIFEST.in enough?

laborleben: and adding '*.v'  after '*.yaml'.

laborleben: Also consider that the answers are more than 2 years old. Quite a bit for setuptools.

themperek: >Isn't MANIFEST.in enough?

As I see it will be part of package but will not be installed during installation.

PR #98, opened 2019-01-08
Increase Python 3 compatibility

DavidLP:
Add partially missing Python 3 compatibility for:
- print function
- int --> long

Other needed changes have been checked for using `python-modernize`.

### DIFF
--- a/basil/HL/MIO_PLL.py
+++ b/basil/HL/MIO_PLL.py
@@ -8,6 +8,7 @@
 import logging
 
 from basil.HL.HardwareLayer import HardwareLayer
+import six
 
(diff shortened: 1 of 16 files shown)


laborleben:
Why we need six at all?
```python
import sys
if sys.version_info > (3,):
    long = int
```
and
```python
list(filter(...))
```
should do the job for Python 2 and 3.

DavidLP:
Six is *the* standard library for python 2 / 3 compatibility. Why not to use it? Also when we drop python 2 support we can spot easily the needed changes.

DavidLP:
BTW.: Introducing `six` is not part of this PR.

PR #100, opened 2019-01-09
Py3k string fixes

DavidLP:
The `socket` library returns `binary arrays` under Python 3 and  `strings` under Python 2. This is treated here.

The `SiTCP` module was tested with hardware using `bdaq53` with Python 2 and Python 3.

The fridges HL was also changed. Not tested but also not really important.


laborleben:
Thanks, this is much cleaner now.

PR #102, opened 2019-01-11
FIX: Fix encoding error in Serial TL due to python3

michaeldaas:
This PR fixes the encoding error in Serial TL that was coming up with the switch to python 3

### DIFF
--- a/basil/TL/Serial.py
+++ b/basil/TL/Serial.py
@@ -40,9 +40,9 @@ class Serial(TransferLayer):
 
     def write(self, data):
         if self.write_termination is None:
-            self._port.write(bytes(data))
+            self._port.write(bytes(data, 'ascii'))
         else:
-            self._port.write(bytes(data + self.write_termination))
+            self._port.write(bytes(data + self.write_termination, 'ascii'))
 
     def read(self, size=None):
         if size is None:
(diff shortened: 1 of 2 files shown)


basil/TL/Serial.py
```diff
@@ -71,4 +71,4 @@ def _readline(self):  # http://stackoverflow.com/questions/16470903/pyserial-2-6
             data += character
             if not character:
                 break
-        return data
+        return data.decode('utf-8')
```
DavidLP: I am not sure this can be done. There might be bytes returned that have to representation in utf-8, no? Did you check this?

basil/TL/Serial.py
```diff
@@ -40,9 +40,9 @@ def close(self):
 
     def write(self, data):
         if self.write_termination is None:
-            self._port.write(bytes(data))
+            self._port.write(bytes(data, 'utf-8'))
```
DavidLP: Why is this needed?

michaeldaas: the constructor `bytes` apparently needs this explicit argument now with python 3. `'utf-8'` should have been the default, so it was always implicitly used

basil/TL/Serial.py
michaeldaas: i only checked with a TTi, a Keithley sand the probestation via serial. Those work fine...

themperek: Seems like utf-8 is a way to go https://pyserial.readthedocs.io/en/latest/pyserial_api.html#serial.Serial.write

@DavidLP We should merge this and see.

Issue #103, opened 2019-01-11
Serial communication much slower with python 3


michaeldaas:
With python 3, a simple measurement with devices using a `Serial` TL are by a factor of 4-5 slower than with python 2.

themperek:
Can you run profiler? 

https://github.com/rkern/line_profiler

michaeldaas:
will do..

DavidLP:
That would be really weird, since the slow part of serial should be the hardware interface

DavidLP:
Is it really slower compared to Python 2 or slower compared to the code before the changes for python 3? Meaning does the code run also slower now under python 2?

michaeldaas:
fixed in development

PR #104, opened 2019-01-14
FIX: Fix wrong parsing of read termination string

michaeldaas:
This PR fixes the 2 second timeout in python3 when reading Serial devices. See #103


michaeldaas:
Should the write termination in line 32 also be cast to `bytes()`?

PR #105, opened 2019-01-18
ENH: encapsulate tobytes call for python 2/3 compatibility

DavidLP:
Otherwise insane amount of:
```
DeprecationWarning: tostring() is deprecated. Use tobytes() instead.
```
warnings

### DIFF
--- a/basil/HL/RegisterHardwareLayer.py
+++ b/basil/HL/RegisterHardwareLayer.py
@@ -12,6 +12,7 @@ import array
 from collections import namedtuple
 from six import integer_types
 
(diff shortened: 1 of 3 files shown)


laborleben:
Can't we do it this way:

```
def tobytes(v):
    try:
        ...
    except ...:
        ...
```

It would be just clearer to read.

laborleben:
I would not increase the version to v3.0.1.dev0. Why would you? We have not even released v3.0.0.

DavidLP:
Since development is already is use

PR #107, opened 2019-02-08
Keep python 2 compatibility in Serial TL

michaeldaas:
With the last PR, python 3 was supported in Serial TL, but it broke python 2 backwards compatibility. This PR restores compatibility with py2 while keeping py3 working as well.

Also cleaned up the TTi mess where it was wrongly named QL335TP instead of QL355TP in some places.

### DIFF
--- a/basil/HL/tti_ql355tp.py
+++ b/basil/HL/tti_ql355tp.py
@@ -12,7 +12,7 @@ from basil.HL.RegisterHardwareLayer import HardwareLayer
 
 
 class ttiQl355tp(HardwareLayer):
-    ''' HL for the TTi QL3335TP.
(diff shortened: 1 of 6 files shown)


laborleben:
Thanks, I'll check that.

laborleben:
The ```bytes()``` function takes at most one argument in Python 2. How does the 'utf-8' argument work?

michaeldaas:
exactly, this is why it throws an error in python 2 which is caught now. in python 3, you need to provide the encoding explicitly.

laborleben:
Ok, thanks for the clarification. I couldn't find any other usage of bytes(). So I assume that we, at least, fixed the bytes() issue.

PR #114, opened 2019-08-15
Python 3 compatibility and clean up

laborleben:
- Python 3 compatibility
- Python 2 compatibility
- Logging/print issues
- Many small fixes
- Code cleanup

Remark:
Removed `from __future__ import print_function`.
Reason: Has no impact on `print()` when using string formatting.
The only difference between Python 2 and Python 3 is when several arguments are given.
But even in this case Python 2 will NOT crash.


basil/HL/julaboF32HD.py
```diff
@@ -5,69 +5,68 @@
...
-        except:
-            print("julabo.get_status() wrong format ".format(repr(ret)))
-            status=-99
-            status_str=ret
-        return status,status_str
-        
+            tmp = ret.split(" ", 1)
+            status = int(tmp[0])
+            status_str = tmp[1:]
+        except IndexError:
```
themperek: Are you sure is this?

basil/HL/tti_ql355tp.py
```diff
@@ -35,7 +35,7 @@ def ask(self, command):
     def read(self):
         ret = self._intf.read()
         if ret[-2:] != "\r\n":
-            print("ttiTp355tp.read() terminator error")
+            print("ttiTp355tp.read() termination error")
```
themperek: Would be nice this would be logger ;-)

basil/utils/sim/SiLibUsb.py
```diff
@@ -85,8 +85,8 @@ def FastBlockWrite(self, data):
         self._iface.send(req)
 
     def WriteI2C(self, address, data):
-        print('SiUSBDevice:WriteI2C', address, data)  # raise NotImplementedError("To be implemented.")
+        print('SiUSBDevice:WriteI2C: {} {}'.format(address, data))
```
themperek: logging?

examples/bdaq/test/test_bdaq53_eth.py
```diff
@@ -68,23 +71,22 @@ def test(self):
                 print(tick)
                 tick_old = tick
 
-            if doprint==True:
+            if doprint is True:
```
themperek: I am not sure about all those `==` to `is` change see: https://stackoverflow.com/questions/15008380/double-equals-vs-is-in-python "1==True returns True, but 1 is True returns False"

examples/lab_devices/julaboF32HD.py.bak
```diff
@@ -0,0 +1,24 @@
+#
```
themperek: What is this file?

basil/HL/FEI4AdapterCard.py
```diff
@@ -210,17 +207,14 @@ def get_id(self):
         ret = self._read_eeprom(self.ID_ADDR, size=calcsize(self.ID_FORMAT))
         return unpack_from(self.ID_FORMAT, ret)[0]
 
-    @abc.abstractmethod
     def _get_adc_value(self, channel):
-        pass
+        raise NotImplementedError("_get_adc_value() not implemented")
```
DavidLP: Much better

basil/HL/tti_ql355tp.py
```diff
@@ -5,7 +5,7 @@
 # ------------------------------------------------------------
 #
 
-from __future__ import print_function
```
DavidLP: You remove Python 2 compatibility or why you do this?

examples/lab_devices/julaboF32HD.py
```diff
@@ -12,8 +12,8 @@
 
 dut = Dut('julaboF32HD.yaml')
 dut.init()
-print("ID", dut["chiller"].get_identifier())
-print("status", dut["chiller"].get_status())
+print("ID: {}".format(dut["chiller"].get_identifier()))
```
DavidLP: Since formatting is touched: we could decide on using single quotes or double quotes everywhere for string keys

basil/dut.py
```diff
@@ -118,13 +118,13 @@ def catch_exception_on_init(mod):
             except NotImplementedError:
                 pass
 
-        for item in itervalues(self._transfer_layer):
+        for item in self._transfer_layer.values():
```
DavidLP: This would degrade Python 2 compatibility, since calls to DUT would be much slower. Big objects would be created on every access to DUT. I would not drop this, just to make it look slightly prettier.

basil/HL/tti_ql355tp.py
laborleben: Agree.

basil/utils/sim/SiLibUsb.py
laborleben: Yes.

examples/lab_devices/julaboF32HD.py.bak
laborleben: Oops.

DavidLP:
How are the two points for python 2 compatibility (print function and dict iterables) addressed? Or do we want to degrade python 2 compatibility with this PR, since support is nevertheless dropped in 1/2 year?

DavidLP:
I add code style checks. OK?

basil/HL/tti_ql355tp.py
themperek: I would not remove for now. It does not cost much to keep it for us.

DavidLP:
Sure, Python 2 will not crash when using parenthesis with print build-in. But it will convert the arguments to a tuple and print the tuple, no? Why would you like to do that?

laborleben:
I don't see any issue with `print()`.
1. The "lists" were replaced by formatted strings.
2. All `print()` statements were removed from user code.
3. `print()` only occur in tests and examples. Code is static and will print nicely in Python 2 and 3.

DavidLP:
Ok, then let's keep the parenthesis in the few examples under Python 2.

DavidLP:
> Reason: Has no impact on print() when using string formatting.

Nice, did not know that!

examples/bdaq/test/test_bdaq53_eth.py
laborleben: Fixed

PR #119, opened 2019-10-04
BUG: Fixed comparator in order to go below default frequency

PatrickAhl:


basil/HL/si570.py
```diff
@@ -52,11 +52,11 @@ def frequency_change(self, freq):  # freq in MHz
 
         new_fdco = freq * HS_DIV * N1
 
-        if 4850.0 < new_fdco > 5670.0:
+        if 4850.0 > new_fdco or new_fdco > 5670.0:
```
themperek: This does not seem to make sense?

PatrickAhl: The calculation below has to be done if it's outside this region.
Page 14: https://www.silabs.com/documents/public/data-sheets/si570.pdf

PatrickAhl: works now for higher and lower frequencies than the default one

PR #120, opened 2019-10-15
WIP: I2C no_stop mode

MarcoVogt:
This MR adds the option to suppress the stop bit at the end of a transmission.
Some chips require multiple write and read commands, separated with only a `start`, but not a `stop` bit.

`set_nostop(bool)` (de)activates this mode. It has to be set before executing the data read/write functions.

To be done: Test/example


MarcoVogt:
Not needed anymore. Closed

PR #122, opened 2019-11-08
Improve bus timing

laborleben:
Improve timing for internal bus by adding additional buffers.


themperek:
We will rather go bit different see in the future: https://github.com/SiLab-Bonn/basil/pull/138 Will removes this tree-state logic making it much faster and better support.
Will close this for now.

PR #126, opened 2020-02-12
Keithley2460

Marrkson:
This PR adds support for the KEITHLEY 2460 and the capability to define the measurement and sourcing range for earlier KEITHLEYS.


laborleben:
Has the identifier really changed to "KEITHLEY INSTRUMENTS,MODEL ..." (without "INC."). What a mess! Hopefully older devices will not change this by a firmware upgrade.

Marrkson:
Yes, the newer models use "KEITHLEY INSTRUMENTS,MODEL ..." (without "INC.")
Buts its great that the rest of the interface is still the same!

PR #127, opened 2020-02-14
ENH: Add channel parameter to sensirion_ekh4

michaeldaas:
This PR adds a `channel` parameter to the thre `get_` methods of the `sensirion_ekh4` HL. This enables using `FunctionalRegisters` to call each sensor connected to the Sensirion Multiplexer box as independent device.

### DIFF
--- a/basil/HL/sensirion_ekh4.py
+++ b/basil/HL/sensirion_ekh4.py
@@ -55,18 +55,21 @@ class sensirionEKH4(HardwareLayer):
                 answer.append(a)
         return answer
 
-    def get_temperature(self, min_val=-40, max_val=200):
+    def get_temperature(self, channel='all', min_val=-40, max_val=200):
         ret = self.ask(r"7e4700b87e")
         values = []
         for j in range(4):
             if ret[2 + 2 * j] == "7f" and ret[2 + 2 * j + 1] == "ff":
                 values.append(None)
             else:
-                values.append(self.cal_ret(
-                    ret[2 + 2 * j] + ret[2 + 2 * j + 1]))
-        return values
+                values.append(self.cal_ret(ret[2 + 2 * j] + ret[2 + 2 * j + 1]))
 
-    def get_humidity(self, min_val=0, max_val=100):
+        if channel == 'all':
+            return values
+        elif isinstance(channel, int):
+            return values[channel]
+
+    def get_humidity(self, channel='all', min_val=0, max_val=100):
         ret = self.ask(r"7e4600b97e")
         values = []
         for j in range(4):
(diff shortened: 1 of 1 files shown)


basil/HL/sensirion_ekh4.py
```diff
@@ -86,7 +93,11 @@ def get_dew_point(self, min_val=-40, max_val=100):
             else:
                 values.append(self.cal_ret(
                     ret[2 + 2 * j] + ret[2 + 2 * j + 1]))
-        return values
+
+        if channel == 'all':
+            return values
+        elif isinstance(channel, int):
+            return values[channel]
```
themperek: Should in return a list even if one channel?

themperek:
Do you really need this? You can just do the same in the user code?

laborleben:
This should do it 👍

michaeldaas:
> Do you really need this? You can just do the same in the user code?

i don't think i can use functional registers in my yaml file to give individual names to the channels without this feature, can i?

basil/HL/sensirion_ekh4.py
michaeldaas: i don't know what's best here?

basil/HL/sensirion_ekh4.py
```diff
@@ -75,9 +78,13 @@ def get_humidity(self, min_val=0, max_val=100):
             else:
                 values.append(self.cal_ret(
                     ret[2 + 2 * j] + ret[2 + 2 * j + 1]))
-        return values
 
-    def get_dew_point(self, min_val=-40, max_val=100):
+        if channel == 'all':
```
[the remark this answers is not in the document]
michaeldaas: done

basil/HL/sensirion_ekh4.py
```diff
@@ -86,7 +91,10 @@ def get_dew_point(self, min_val=-40, max_val=100):
             else:
                 values.append(self.cal_ret(
                     ret[2 + 2 * j] + ret[2 + 2 * j + 1]))
-        return values
+
+        if channelis None:
```
DavidLP: Well, broken pychecker?

michaeldaas: triggerhappy commit-fingers :see_no_evil:

basil/HL/sensirion_ekh4.py
DavidLP: Would return number or list. IF you return always list one cannot use return value directly and this feature would be useless, to my mind.

themperek: Fine with me anyway if usefull.

michaeldaas:
any more changes necessary?

PR #128, opened 2020-02-17
Improving effiency of JTAG data transfert

VincentGuerard:
Improving JTAG effiency by writting 128 bits at a time with basil bus instead of one bit at a time.


basil/HL/JtagMaster.py
```diff
@@ -0,0 +1,287 @@
...
+        """
+        self.WORD_COUNT = value
+
+    def get_word_count(self):
+        """
+        Gets Number of word of mem size to send
+        """
+        return self.WORD_COUNT
+
+    def set_operation(self, value):
```
themperek: Would no be better 
```python 
set_operation("INSTRUCTION")
set_operation("DATA")
```
or similar?

Just magic numbers like 0 and 1 seem bad.

laborleben:
Thanks for PR.

Do you have the module already in use for any kind of project?

basil/HL/JtagMaster.py
```diff
@@ -0,0 +1,287 @@
...
+            )
+
+        bit_number = sum(len(x) for x in data)
+        if bit_number < self._mem_bytes * 8:
+            self.set_size(bit_number)
+        else:
+            raise ValueError(
+                "Size is too big for memory: got %d and memory is: %d" % bit_number,
+                self._mem_bytes * 8,
+            )
```
themperek: Looks like repeating code like in `scan_ir`? Maybe function?

basil/firmware/modules/jtag_master/blk_mem_gen_8_to_1_2k.v
```diff
@@ -0,0 +1,44 @@
+/**
```
themperek: Is the same as for SPI so maybe there from there and not copy

basil/firmware/modules/jtag_master/jtag_master_core.v
```diff
@@ -0,0 +1,478 @@
...
+    .WEB(SEN_INT),
+    .ADDRA(ADDRA_MIN),
+    .ADDRB(ADDRB_MIN),
+    .DINA(BUS_DATA_IN_IB),
+    .DINB(TDO)
+);
+///
+
+// JTAG Master Machine state //
+enum {TEST_LOGIC_RESET,
```
themperek: Check if this synthesise with Vivado.

basil/HL/JtagMaster.py
themperek: maybe `set_command()` ?

basil/HL/JtagMaster.py
VincentGuerard: yes, no problems

basil/firmware/modules/jtag_master/blk_mem_gen_8_to_1_2k.v
VincentGuerard: Ok

basil/firmware/modules/jtag_master/jtag_master_core.v
VincentGuerard: You're right, it does not synthesise with vivado, I will make the changes necessary.

basil/HL/JtagMaster.py
```diff
@@ -0,0 +1,288 @@
...
+        """
+
+        bit_number = self._test_input(data)
+        self.set_size(bit_number)
+
+        data_byte = self._bitlogic2bytes(data)
+        self.set_data(data_byte[::-1])
+
+        self.set_word_count(1)
+        self.set_command(self.jtag_command["INSTRUCTION"])
```
themperek: I think it should be:
```
self.set_command("INSTRUCTION")
```
and conversion inside set_command.

basil/HL/JtagMaster.py
```diff
@@ -93,17 +93,21 @@ def get_word_count(self):
...
     def set_command(self, value):
         """
-        Operation identifier
         IR_SCAN or DR_SCAN
         """
-        self.OPERATION = value
+        if value == "INSTRUCTION":
+            self.OPERATION = self.jtag_command["INSTRUCTION"]
+        elif value == "DATA":
+            self.OPERATION = self.jtag_command["DATA"]
```
themperek: `self.OPERATION = self.jtag_command[value]` ?

VincentGuerard: You're right, that would make more sense

themperek:
Look OK.

Should I marge?

themperek:
I hope 128 is no the only size ;-)

laborleben:
There are still open issues.

laborleben:
I really don't see the use case for all the set_*/get_* function. They are **deprecated**. We have it already in RegisterHardwareLayer generated (with uppercase).

themperek:
Yes. For simple assignment like `self.WORD_COUNT = value`  just use directly.

@laborleben We could even patch this to be able to use lowercase.

laborleben:
Then you change the rules again.

laborleben:
We have "self.set_WORD_COUNT(value)" already.

In fact we have:
self.dut['test_register'].REGISTER = 42
self.dut['test_register']['REGISTER'] = 42
self.dut['test_register'].set_ REGISTER(42)

They are tested and used everywhere.

Not sufficient?

themperek:
@VincentGuerard  Thank you!

PR #129, opened 2020-02-27
BUG: Problems occurs when performing a reset with a slower JTAG clock

VincentGuerard:
The fix to the problem was to keep the module busy while performing a reset.


themperek:
You need to rebase. Or make a new fork from development and apply the patch on top ;-)

PR #130, opened 2020-02-27
 BUG: Fix the reset function of the module. 

VincentGuerard:
A problem occur when slowing down the JTAG clock if the module is not kept busy.


themperek:
@VincentGuerard can you fix this `E128 @ 172:17 - continuation line under-indented for visual indent`

I was to fast.

PR #132, opened 2020-03-09
BUG: Fix memory address not increasing when sending multiple words

VincentGuerard:
I have also added a new test that write data directly into module memory without using the scan_dr function to test multiple word writing.

### DIFF
--- a/basil/HL/JtagMaster.py
+++ b/basil/HL/JtagMaster.py
@@ -8,6 +8,8 @@ import struct
 
 from basil.HL.RegisterHardwareLayer import RegisterHardwareLayer
 from basil.utils.BitLogic import BitLogic
+from bitarray import bitarray
+import numpy as np
 
 
 class JtagMaster(RegisterHardwareLayer):
@@ -102,7 +104,7 @@ class JtagMaster(RegisterHardwareLayer):
         self.SIZE = bit_number
 
         data_byte = self._bitlogic2bytes(data)
-        self.set_data(data_byte[::-1])
+        self.set_data(data_byte)
 
         self.WORD_COUNT = 1
         self.set_command("INSTRUCTION")
@@ -111,8 +113,7 @@ class JtagMaster(RegisterHardwareLayer):
             pass
 
         received_data = self.get_data(size=len(data_byte))
-        ret = self._bytes2bitlogic(received_data, bit_number)
-        rlist = self._split_bitlogic(ret, data)
+        rlist = self._bytes2bitlogic(received_data, bit_number, data)
 
         return rlist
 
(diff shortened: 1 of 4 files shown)


tests/test_SimJtagMaster.py
```diff
@@ -395,16 +390,55 @@ def jtag_tests(self):
...
+
+        self.chip["JTAG"].scan_ir([DEBUG, DEBUG])
+
+        # The ring register (DEBUG register) is 32 bits long, so the data have to be arranged like this : [WORD1(dev1) WORD1(dev2) WORD2(dev1) WORD2(dev2) ...]
+        ba = BitLogic()
+        all_data = bytearray()
+        for word in range(word_count):
+            ba.frombytes(data[word * word_size_byte:(word + 1) * word_size_byte].tobytes())
+            ba.fill()
+            ba.reverse()
```
themperek: Why you need BitLogic here? Can you just `self.chip["JTAG"].set_data(data)`?

Why we have to reverse here?

VincentGuerard: I use the Bitlogic only to reverse the data. We need to reverse because I write directly into the memory and the MSB and LSB is reversed in hardware right? Like in Verilog we have reg [32:0] and in python we have reg [0:32].
Do you know a better way to achieve this ?

themperek: We should then in hardware go from the back for the word?

VincentGuerard: You mean instead of reversing the word in python, we should read in reverse in the hardware ?

themperek: Yes.

Here: https://github.com/SiLab-Bonn/basil/blob/6d72e7464a2cccb67eef5ca640748ce6f47fb2ad/basil/firmware/modules/jtag_master/jtag_master_core.v#L168

Somethng like:
`assign memout_addrb = (out_word_cnt * CONF_BIT_OUT) + CONF_BIT_OUT - 1 -out_bit_cnt;`

?

basil/HL/JtagMaster.py
```diff
@@ -128,14 +122,17 @@ def scan_dr(self, data, words=1):
         Data must be a list of BitLogic or string of raw data
         """
 
-        bit_number = self._test_input(data)
-        self.SIZE = bit_number
+        bit_number = self._test_input(data, words)
+        if words != 1:
```
themperek: Why you need this `if`?

basil/HL/JtagMaster.py
```diff
@@ -128,14 +122,17 @@ def scan_dr(self, data, words=1):
...
+            self.SIZE = int(bit_number / words)
+        else:
+            self.SIZE = bit_number
```
themperek:
```suggestion
        self.SIZE = int(bit_number / words)
```

themperek:
@VincentGuerard Thank you for your work. In the current situation I would marge this and see later.

themperek:
Thank you @VincentGuerard !

VincentGuerard:
You're welcome !

PR #135, opened 2020-10-31
Enhance temperature control and measurement

michaeldaas:
This MR adds more functions to the new Weiss LabEvent HL. Sorry i did not include this in the original MR but i just discovered another project that "documented" all possible function definitions.

Furthermore, this MR tries to fix the remaining problems with the Sensirion EKH4 HL. While the readout crashing due to random buffer was fixed already with the previous MR, the sensor still reports wrong / no values under certain conditions. As a workaround, the implementation of a sanity check of the results is finished.

I still have hopes of understanding / fixing the underlying issue though...

### DIFF
--- a/basil/HL/sensirion_ekh4.py
+++ b/basil/HL/sensirion_ekh4.py
@@ -63,7 +63,11 @@ class sensirionEKH4(HardwareLayer):
             if ret[2 + 2 * j] == "7f" and ret[2 + 2 * j + 1] == "ff":
                 values.append(None)
             else:
-                values.append(self.cal_ret(ret[2 + 2 * j] + ret[2 + 2 * j + 1]))
+                val = self.cal_ret(ret[2 + 2 * j] + ret[2 + 2 * j + 1])
+                if val < min_val or val > max_val:
+                    values.append(None)
+                else:
+                    values.append(val)
 
         if channel is None:
             return values
@@ -76,7 +80,11 @@ class sensirionEKH4(HardwareLayer):
             if ret[2 + 2 * j] == "7f" and ret[2 + 2 * j + 1] == "ff":
                 values.append(None)
             else:
-                values.append(self.cal_ret(ret[2 + 2 * j] + ret[2 + 2 * j + 1]))
+                val = self.cal_ret(ret[2 + 2 * j] + ret[2 + 2 * j + 1])
+                if val < min_val or val > max_val:
+                    values.append(None)
+                else:
+                    values.append(val)
 
         if channel is None:
             return values
@@ -89,7 +97,11 @@ class sensirionEKH4(HardwareLayer):
             if ret[2 + 2 * j] == "7f" and ret[2 + 2 * j + 1] == "ff":
                 values.append(None)
             else:
-                values.append(self.cal_ret(ret[2 + 2 * j] + ret[2 + 2 * j + 1]))
+                val = self.cal_ret(ret[2 + 2 * j] + ret[2 + 2 * j + 1])
+                if val < min_val or val > max_val:
+                    values.append(None)
+                else:
+                    values.append(val)
 
         if channel is None:
             return values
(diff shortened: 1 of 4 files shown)


michaeldaas:
From myobservations, it seems the sensors report the wrong temperature only between +43.52°C and +46.53°C as well as between +48.57°C and +51.20°C. Additionally, the sensor reports wrong data when the temperature approaches -40°C, but to be fair, this is the minimum of the rated temeprature range for the SHT35 sensor in use..

themperek (approved):
For me, ok but maybe @DavidLP should check.
Maybe add this comment to the file about temperature or add some warning.

basil/HL/sensirion_ekh4.py
```diff
@@ -76,7 +80,11 @@ def get_humidity(self, channel=None, min_val=0, max_val=100):
             if ret[2 + 2 * j] == "7f" and ret[2 + 2 * j + 1] == "ff":
                 values.append(None)
             else:
-                values.append(self.cal_ret(ret[2 + 2 * j] + ret[2 + 2 * j + 1]))
+                val = self.cal_ret(ret[2 + 2 * j] + ret[2 + 2 * j + 1])
+                if val < min_val or val > max_val:
```
DavidLP, line 84: So, you would like to re-implement the feature of checking the return value, that you have deleted before?

basil/HL/weiss_labevent.py
```diff
@@ -22,12 +23,23 @@ class weissLabEvent(HardwareLayer):
...
+
     def __init__(self, intf, conf):
         super(weissLabEvent, self).__init__(intf, conf)
 
     def init(self):
         super(weissLabEvent, self).init()
 
+        info = self.get_info()
+        if info != 'LabEvent T/210/70/5':
+            logging.error('Not the expected climatechamber! Expected {0}, got {1}'.format('LabEvent T/210/70/5', info))
```
DavidLP, line 41: In basil we define a logger per file at the beginning of the module
```python
import logging
logger = logging.getLogger(__name__)
```

DavidLP: Otherwise we cannot filter logs efficiently in the DAQs that use basil. See [here](https://github.com/SiLab-Bonn/basil/pull/79)

michaeldaas:
chill guys, this is WIP.. i was still hoping to fix the underlying issue with the Sensirion EK-H4, but seems like i don't manage. Since it's EOL anyways, i will push a nicer qorkaround tonight and will leave it at that..

basil/HL/sensirion_ekh4.py
michaeldaas: it was never implemented, only the parameters were there but no check. i have a cleaner workaround now, will push tonight

DavidLP:
Oh, yea. I just saw. Was only reading Tomeks invitation to review. hehe

themperek:
` michaeldaas requested review from themperek and DavidLP 2 days ago` do not do this if not ready .... ;-)

DavidLP:
:-D

basil/HL/sensirion_ekh4.py
DavidLP: I implemented it in my original version 5 years ago. But you did not delete it, [Toko did](https://github.com/SiLab-Bonn/basil/commit/fa4f9f82fe6ce12a2a744b5fcc875ddc36e8d07f).

michaeldaas: some error correction and more secure handling of corrupt data now implemented, so this workaround was removed completely

basil/HL/weiss_labevent.py
michaeldaas: done

michaeldaas:
sorry, my bad.. i'm too used to gitlab..

michaeldaas:
i implemented some error handling for sensirion ekh4, but there are some cases where it just injects random bytes into the data.. these cases are now reliably set to none, as far as i can tell from testing in the chamber. no chance of reconstructing the values unfortunately..

michaeldaas:
now i'm done and would like a review, please ;)

michaeldaas:
you want to review again @DavidLP or should i just merge?

PR #136, opened 2020-11-03
Driver for the sensirion sht85 sensor

konstantinmauer:
Adds a driver for the sensirion sht85 sensor and a sensirion sensor bridge. It uses official drivers of sensirion to communicate with the sensor bridge.


basil/HL/SensirionBridgeDevice.py
```diff
@@ -0,0 +1,53 @@
+# ------------------------------------------------------------
+# Copyright (c) All rights reserved
+# SiLab, Institute of Physics, University of Bonn
+# ------------------------------------------------------------
+#
+
+from basil.HL.RegisterHardwareLayer import HardwareLayer
+
+class SensirionBridgeI2CDevice(HardwareLayer):
```
michaeldaas, line 9: Is this additional HL actually needed? As far as i can tell, the commands here control the connected sensor (directly or indirectly) and could be integrated in the corresponding HL. I'm not sure we will ever use something other than the T / H  sensors we already have with this bridge...

basil/HL/sensirion_sht85.py
```diff
@@ -0,0 +1,179 @@
...
+    '''
+
+    def __init__(self, intf, conf):
+        super(sensirionSHT85, self).__init__(intf, conf)
+
+    def init(self):
+        super(sensirionSHT85, self).init(0x44)
+
+        try:
+            import crcmod
```
michaeldaas, line 35: why import in `init`?

basil/HL/sensirion_sht85.py
```diff
@@ -0,0 +1,179 @@
...
+            self.crc_func = crcmod.mkCrcFun(0x131, initCrc=0xFF, rev=False, xorOut=0x00)
+        except ModuleNotFoundError:
+            logger.warning("You have to install the package 'crcmod'! Transmission errors will not be catched.")
```
michaeldaas, line 38: 'catched' -> 'caught'
Why masking the import error here anyways? Either the dependency is needed or not..

basil/HL/sensirion_sht85.py
```diff
@@ -0,0 +1,179 @@
...
+            self.crc_func = lambda x: 0
+        
+        if 'repeatability' in self._init.keys():
```
michaeldaas, line 41: this is usually done with `repeatability = self._init.get('repeatability', 'low')`

basil/HL/sensirion_sht85.py
```diff
@@ -0,0 +1,179 @@
...
+    def enable_heater(self):
+        self._write([0x30, 0x6D])
+        
+    def disable_heater(self):
+        self._write([0x30, 0x66])
+
+    def is_heater_on(self):
+        return self._get_status()[0]&(1<<13) > 0
+        
+    def reset_sensor(self):
```
michaeldaas, line 156: Maybe you could add some documentation what whis command actually resets?

basil/HL/sensirion_sht85.py
```diff
@@ -0,0 +1,179 @@
...
+        self._write([0x30, 0xA2])
+
+    def get_repeatabilaty(self, repeatability=None):
```
michaeldaas, line 159: this seems unnecessary..

basil/HL/sensirion_sht85.py
```diff
@@ -0,0 +1,179 @@
...
+        if repeatability is None:
+            return self.repeatability
+        else:
+            return repeatability
+
+    def to_temperature(self, data):
```
michaeldaas, line 165: shuold be `_to_temperature()`

basil/HL/sensirion_sht85.py
```diff
@@ -0,0 +1,179 @@
...
+        return -45+175*(data[0]/(2**16-1))
+
+    def to_humidity(self, data):
```
michaeldaas, line 168: should be `_to_humidity()`

basil/HL/sensirion_sht85.py
```diff
@@ -0,0 +1,179 @@
...
+        return 100*(data[1]/(2**16-1))
+
+    def to_dew_point(self, T, RH):
```
michaeldaas, line 171: the sensor doesn't support dew point measurement? I'm not sure if this kind of calculation should be handled by basil.. on the other hand it would be nice to support the same function as other temperature sensors, so they can be interchanged...

basil/TL/SensirionSensorBridge.py
```diff
@@ -0,0 +1,90 @@
...
+logger = logging.getLogger(__name__)
+
+try:
+    from sensirion_shdlc_driver import ShdlcSerialPort, ShdlcConnection
+    from sensirion_shdlc_sensorbridge import SensorBridgePort, SensorBridgeShdlcDevice
+    from sensirion_shdlc_sensorbridge.device_errors import SensorBridgeI2cTimeoutError
+    TimeoutError = SensorBridgeI2cTimeoutError
+except ModuleNotFoundError:
+    logger.error("You have to install the package 'sensirion_shdlc_sensorbridge' to use this device.")
+    raise
```
michaeldaas, line 22: Why are you handling import errors manually?

basil/TL/SensirionSensorBridge.py
```diff
@@ -0,0 +1,90 @@
...
+
+bridge_ports = {
```
michaeldaas, line 24: i guess this should go directly in the class definition?

basil/TL/SensirionSensorBridge.py
```diff
@@ -0,0 +1,90 @@
...
+    The Sensirion Sensor Bridge is connected via USB and allows communication to two I2C ports.
+    '''
+
+    def __init__(self, conf):
+        super(SensirionSensorBridge, self).__init__(conf)
+
+    def init(self):
+        super(SensirionSensorBridge, self).init()
+
+        if 'port' in self._init.keys():
```
michaeldaas, line 42: se above: use `self._intf.get()`

basil/TL/SensirionSensorBridge.py
```diff
@@ -0,0 +1,90 @@
...
+            self.port = self._init['port']
+        else:
+            self.port = '/dev/ttyUSB0'
```
michaeldaas, line 45: i think we never define a default port with the other TLs.. makes no sence to want to use some hardware nad not supply the port manually...

basil/TL/SensirionSensorBridge.py
```diff
@@ -0,0 +1,90 @@
...
+        if 'baudrate' in self._init.keys():
+            self.baudrate = self._init['baudrate']
+        else:
+            self.baudrate = 460800
+
+        self.ser = ShdlcSerialPort(port=self.port, baudrate=self.baudrate)
+
+    def get_sensor_bridge_port(self, port):
```
michaeldaas, line 53: again redundant. if the dict is in the class definition, you can directly access it from the outside without a getter..

examples/lab_devices/sensirionSHT85.py
```diff
@@ -0,0 +1,33 @@
...
+#
+
+from basil.dut import Dut
+
+import time
+
+dut = Dut('examples/lab_devices/sensirionSHT85.yaml')
+dut.init()
+dut['Thermohygrometer'].printInformation()
+print(dut['Thermohygrometer'].get_temperature_and_humidity())
```
michaeldaas, line 15: This can be added to the temperature control example file.. maybe we split this into 'temperature_sensors' and 'temperature_control' now, since it's getting quite big..

michaeldaas: actually, i will already implement this with #135 so you can add your code to the 'temperature_sensors' example file once it's merged..

michaeldaas (commented):
this thing is getting quite complicated for a temperature sensor 🙈 
would like to hear what @DavidLP thinks about some of my comments..

DavidLP:
Pretty nice first commit into basil!

DavidLP:
I do not know the Sensirion box etc. so it is tough for me to judge this. From what I read, it works as the following:
- there is a sensirion box that allows to have up to 2 ic2 devices connected to it. This is the TL `SensirionSensorBridge`
- the i2 devices can be connected to it. A generic i2c device is `SensirionBridgeI2CDevice`
- a specific i2c device are the temperature devices `sensirionSHT85`

I mean that is (as I can quickly judge) a very clean and generic implementation to support hardware we have and to support hardware we _don't_ have ;-)

If we are only going to use the bridge with the temperature sensors, one could maybe just make one driver and skip the extra TL.
For lab devices I originally used the TL to specify the means of communication (RS232, TCP/IP, GPIB, USB and the library NI VISA, python.socket, ...). This makes a lot of sense for lab devices that have multiple interfaces (e.g. TTi, Keithleys, ...). I am not sure what we gain for having such generic TL/HL pattern, when the TL is not interchangeable, not configurable and fixed (@michaeldaas an me discussed that recently, too).

basil/HL/SensirionBridgeDevice.py
konstantinmauer: I just thought, that this way more sensors could be added. But if we will not use any other sensors, having only the one file would be simpler.

basil/HL/sensirion_sht85.py
konstantinmauer: That way the dependency is not required. But checking for transmission errors is always prefered I think.

basil/HL/SensirionBridgeDevice.py
michaeldaas: mhh.. what other sensors does this bridge support at all? sht35 is basically the same? (apart from the heater, obviously).. maybe it makes sense to also support this. the sht35 that we have from the old ek-h4 should work with this bridge as well, correct?

basil/HL/sensirion_sht85.py
konstantinmauer: Unfortunately not. I added it here to suport at least some dew point measurement, since the bdaq53 periphery class defined a getter for the dew point, so it is at least kompatible.

basil/TL/SensirionSensorBridge.py
konstantinmauer: I did not want to have this pakage as a dependecy for basil. I just wanted to message which package to install if required.

basil/HL/SensirionBridgeDevice.py
konstantinmauer: No it does not. It has its own protocol, that is simillar to I2C but not suported by the bridge. The bridge can be used to communicate arbitrary I2C or SPI devices.

michaeldaas: ok, i guess in this case the general implementation makes sense. but that's a TL in the strict definition. why do you need a standalone HL?

basil/HL/sensirion_sht85.py
michaeldaas: mhh, not the best style, though. what do you think @DavidLP ?

basil/HL/SensirionBridgeDevice.py
konstantinmauer: My thought behind this seperation was: 
- The bridge enables cummunication to devices of multible types (currently just I2C). So its defined in the TL.
- Multiple devices can be connected to the bridge. They are defined in the HL.
- Information like the bridge port or the I2C adress have to be stored per device and are therefore also part of the HL. But since they are not specific to the device type, I defined a generic I2C device, that stores this kind of information and handels all that is specific to the bridge.
- A specific device inherits from this class and can just use the more generic functions without having to know how the bridge works

I hope this makes sense

konstantinmauer: But you are right. Technicly this should be TL, but than we would need to have two TL, one for the bridge and anover for each device. I'm not sure, how to implement this correctly.

michaeldaas: i see. another layer of base class.. ok, i guess. if there's enough generic stuff to justify this, why not.

examples/lab_devices/sensirionSHT85.py
michaeldaas: #135 is merged, so you can rebase and add your code to the 'termperature_sensors' example

konstantinmauer: Done

basil/HL/sensirion_sht85.py
konstantinmauer: I think, this is the last thing open. I could place this if condition ouside the class definition. Would that be better?

michaeldaas: i guess so, but i would waid for @DavidLP 's opinion

basil/TL/SensirionSensorBridge.py
```diff
@@ -0,0 +1,78 @@
...
+import logging
+import array
+
+from basil.TL.TransferLayer import TransferLayer
+
+logger = logging.getLogger(__name__)
+
+from sensirion_shdlc_driver import ShdlcSerialPort, ShdlcConnection
+from sensirion_shdlc_sensorbridge import SensorBridgePort, SensorBridgeShdlcDevice
+from sensirion_shdlc_sensorbridge.device_errors import SensorBridgeI2cTimeoutError
```
themperek, line 17: You could try to catch import error and give a link to install.

michaeldaas: its on pypi. https://pypi.org/project/sensirion-shdlc-driver/
should be a prerequisit to basil and added to `setup.py`?

themperek: Probably not since most people will not sue it. Not very strong opinion.

basil/HL/sensirion_sht85.py
DavidLP: fine in init

basil/TL/SensirionSensorBridge.py
DavidLP: Should not be in setup.p; otherwise we clutter it too much with dependencies.

basil/HL/sensirion_sht85.py
DavidLP: the you can use a yaml describing your setup (even with this sensor in the yaml), that works as long as you do not call init. Fine with me here

DavidLP:
I recommend to configure your editor in a way that it takes care of the code style. Good editors (pydev, VStudio Code, ...) can do that. When you clean up the formatting issues https://travis-ci.org/github/SiLab-Bonn/basil/jobs/741928925#L5989 it is fine to be merged. You can check for code flaws by typing:
```
flake8-diff origin/development
```

DavidLP:
Also, you see that the CI failed due to code flaws

PR #139, opened 2020-12-01
Force PyVisa-sim to version 0.3 (fix ci for 2.7)

themperek:


DavidLP:
I thought we wanted to drop python 2.7 support?

themperek:
> I thought we wanted to drop python 2.7 support?

Yes. This is just for the last check.

themperek:
Now it fails 3.7 so one ould need an if. Ok I gonore this.

PR #141, opened 2020-12-02
GHA integration

themperek:
- remove support for python 2.7
- fix lint-flake8 errors
- add Github Actions integration
- remove Travis integration
- update badges


themperek:
@DavidLP Ready for review.

DavidLP:
Uhhh, so fancy. Will have a closer look likely tomorrw

.github/workflows/regression-tests.yml
```diff
@@ -0,0 +1,45 @@
+name: Regression Tests
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    name: Python ${{matrix.python-version}}
+    runs-on: ubuntu-20.04
```
DavidLP, line 27: We could also run on mac OS and Windows, no? ;-) Maybe a good idea, since there are some windows MacOS users from time to time

DavidLP (approved):
Looks nice. We could also run on Windows/Mac but just changing 2 lines of code? Its for free for us?

DavidLP:
We still need [coveralls](https://github.com/marketplace/actions/coveralls-github-action) running
```yaml
- name: Coveralls
      uses: coverallsapp/github-action@master
      with:
        github-token: ${{ secrets.GITHUB_TOKEN }}
```

.github/workflows/regression-tests.yml
themperek: We can although not sure if needed.

DavidLP (changes requested):
Coveralls missing

DavidLP:
Lets switch to codecov. It's the new one everybody uses

.github/workflows/regression-tests.yml
DavidLP: let's do it in other PR, if wanted

DavidLP (approved):
Nice. Fine to merge

PR #147, opened 2020-12-04
Fix bug and compability issues of SHT85 driver

konstantinmauer:
A bug caused, that, with each call of get_temperaruture(), three measurements where performed by the sensor, while only the result with the chosen repeatablilaty was returned and the other two results where discarded. Also the decoding of the response was not compatible with python 2.

### DIFF
--- a/basil/HL/sensirion_sht85.py
+++ b/basil/HL/sensirion_sht85.py
@@ -162,11 +162,12 @@ class sensirionSHT85(SensirionBridgeI2CDevice):
         return 100 * (float(data[1]) / (2**16 - 1))
 
     def to_dew_point(self, T, RH):
-        if T < 0:
-            T_n = 243.12
-            m = 17.6
-        else:
-            T_n = 272.62
-            m = 22.46
-        a = np.log(RH / 100.0) + m * T / (T_n + T)
-        return T_n * a / (m - a)
+        ''' returns the dew point using an approximation
+            approximation specified by Sensirion:
+            http://irtfweb.ifa.hawaii.edu/~tcs3/tcs3/Misc/Dewpoint_Calculation_Humidity_Sensor_E.pdf
+        '''
+        if RH == 0:
+            RH = self._to_humidity((0, 1))  # lowest non-zero rel. humidity
+        H = (np.log10(RH) - 2) / 0.4343 + (17.62 * T) / (243.12 + T)
+        Dp = 243.12 * H / (17.62 - H)
+        return Dp
--- a/basil/HL/sensirion_sht85.py
+++ b/basil/HL/sensirion_sht85.py
@@ -6,6 +6,7 @@
 
 import numpy as np
 import logging
+import struct
 
 from basil.HL.SensirionBridgeDevice import SensirionBridgeI2CDevice
 
@@ -49,7 +50,7 @@ class sensirionSHT85(SensirionBridgeI2CDevice):
                 if self.crc_func(rx_data[i * 3:(i + 1) * 3]):
                     break
                 else:
-                    data[i] = int.from_bytes(rx_data[i * 3: i * 3 + 2], byteorder='big')
+                    data[i] = struct.unpack('>H', rx_data[i * 3: i * 3 + 2])[0]
             else:
                 return data
             continue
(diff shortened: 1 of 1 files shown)


basil/HL/sensirion_sht85.py
```diff
@@ -6,17 +6,19 @@
 
 import numpy as np
 import logging
+import struct
 
 from basil.HL.SensirionBridgeDevice import SensirionBridgeI2CDevice
 
 logger = logging.getLogger(__name__)
+logging.getLogger("sensirion_shdlc_driver.connection").setLevel(logging.ERROR)
```
michaeldaas, line 14: this seems like a workaround for a problem in the driver? why is it necessary to do this here?

basil/HL/sensirion_sht85.py
```diff
@@ -6,17 +6,19 @@
...
 
 
 class sensirionSHT85(SensirionBridgeI2CDevice):
     '''
     Driver for the Sensirion SHT85 temperature and humidity sensor.
     Measurements can be performed in three repeatability modes:
-    low (0.15°C 0.21%RH) (default), medium (0.08°C 0.15%RH), high (0.04°C 0.08%RH)
+    low (0.15C 0.21%RH) (default), medium (0.08C 0.15%RH), high (0.04C 0.08%RH)
```
michaeldaas, line 21: is there an actual problem with the `°` sign? if have so far not managed to find any circumstance that produced problems with this?

basil/HL/sensirion_sht85.py
```diff
@@ -59,12 +61,12 @@ def _write(self, command):
...
-            "low": self._read([0x24, 0x16], read_n_words=read_n_words, timeout_us=4500),
-            "medium": self._read([0x24, 0x0B], read_n_words=read_n_words, timeout_us=6500),
-            "high": self._read([0x24, 0x00], read_n_words=read_n_words, timeout_us=15500),
+        params = {
+            "low": ([0x24, 0x16], 4500),
+            "medium": ([0x24, 0x0B], 6500),
+            "high": ([0x24, 0x00], 15500),
         }[self.repeatability]
-        return data
+        return self._read(params[0], read_n_words=read_n_words, timeout_us=params[1])
```
michaeldaas, line 69: ok, nice

michaeldaas (changes requested):
ok, we should commit the bug fix. The dew point calculation also seems nicer.
but what about the two first changes? is this really necessary?

basil/HL/sensirion_sht85.py
konstantinmauer: When, data should be read but no data is received before the timeout, the sensirion_shdlc_driver writes a warning and throws a Exception. But in case of the asynchronous read we do not want to wait if no data is available and want to catch the Exception. I thought, in this case, we also do not want to print any warning. But maybe one can only ignore warings, if one is in the asynchronous mode?

basil/HL/sensirion_sht85.py
konstantinmauer: I thing python was complaining on a Windows machine some time ago. At least someone told me.  I did not check again myself, but I can do that.

basil/HL/sensirion_sht85.py
michaeldaas: mhh, ok. a bit hacky but fine for me i guess..

basil/HL/sensirion_sht85.py
michaeldaas: afaik, `cmd` will not display the symbol but it should not crash. making it uglier for everyone just so windows `cmd` doesn't complain isn't fair to my mind ;)

konstantinmauer: Okay... and even if a Windows user experiences some error, he should be capable of removing the ° signs by himself

michaeldaas: if it actually crashes because of this, it's worth to remove it i guess. would be nice if you could actually check ;)

konstantinmauer: It does not complain on my Windows notebook. Since I do not remember the exact Error someone showed me before, I would  close this, unless it comes up again.

michaeldaas (approved):
thanks for the changes! fine for me now.

DavidLP (approved):
Do not understand everything; but looks good.

PR #151, opened 2020-12-19
Add more split bus tests

flooklab:


tests/test_SimFifo8to32_sbus_top.v
```diff
@@ -0,0 +1,109 @@
+/**
```
themperek, line 1: Would not be nicer to have one Verilog module and use `define` to switch buses?

themperek: Less repeating code.

flooklab: Some like this? a6d0d9ad7eb7c9b327020c85d6789a7cde337b55

flooklab: Can we define macros through cocotb for different simulators in a consistent way?

tests/test_SimGpio.v
```diff
@@ -7,15 +7,31 @@
```
themperek, line 7: Something like this :+1:
 Was thinking to do this only on Verilog that has this simple sbus trick but this is also OK.
Can you do this on all other files?

flooklab: Yes.
Can we set the macros also for iverilog so that CI passes?
Or just skip the sbus tests?

themperek: `extra_defines` in `cocotb_makefile` ?:
https://github.com/SiLab-Bonn/basil/blob/a04bc83ac9fa0b9324f01144accb4f11df613807/basil/utils/sim/utils.py#L17-L18

flooklab: Have to use `compile_args` to use with both icarus and verilator.
I will modify the other tests.

themperek: Probably better fix in `basil/basil/utils/sim/utils.py` and use `extra_defines` if needed.

themperek:
@flooklab Thank you!

PR #153, opened 2020-12-23
Use pytest markers for CI to select only sbus-based tests to run with Verilator

flooklab:
Could run Verilator-compatible `sbus` versions of tests with Verilator using pytest:
`SIM=verilator pytest -m verilator tests/test_*.py`


themperek:
Have to add `pytest` here:
https://github.com/SiLab-Bonn/basil/blob/6733132b0b7ee5e08bdac3c8c9df6f4f7bdcbdcf/.github/workflows/regression-tests.yml#L55

Maybe better use `ventilator` as a mark?

themperek:
Thank you.

PR #158, opened 2021-05-03
Signatone support

Marrkson:
This PR adds support for the signatone prober in paris.


basil/TL/Visa.py
```diff
@@ -32,10 +34,13 @@ def init(self):
         backend = self._init.get('backend', '')  # Empty string means std. backend (NI VISA)
         rm = visa.ResourceManager(backend)
         try:
-            logger.info('BASIL VISA TL with %s backend found the following devices: %s', backend, ", ".join(rm.list_resources()))
+            logger.info('BASIL VISA TL with %s backend found the following devices: %s', backend,
+                        ", ".join(rm.list_resources()))
```
michaeldaas, line 38: this change is due to pep8 E501 enforcement? what is the policy in basil? i don't find it useful in modern times with widescreen displays everywhere...

basil/TL/Visa.py
```diff
@@ -32,10 +34,13 @@ def init(self):
...
         except NotImplementedError:  # some backends do not always implement the list_resources function
             logger.info('BASIL VISA TL with %s backend', backend)
-        self._resource = rm.open_resource(**{key: value for key, value in self._init.items() if key not in ("backend",)})
+
+        self._resource = rm.open_resource(
+            **{key: value for key, value in self._init.items() if key not in ("backend",)})
```
michaeldaas, line 43: same as above

examples/lab_devices/SignatoneProber.yaml
```diff
@@ -0,0 +1,13 @@
+transfer_layer:
```
michaeldaas, line 1: maybe you also want to add a (commented) example to the `probestation.py` file to make the example more obvious?

basil/TL/Visa.py
```diff
@@ -45,7 +50,27 @@ def write(self, data):
         self._resource.write(data)
 
     def read(self):
-        self._resource.read()
+        if self._resource.read_termination == "":
+            ret = ""
+            while True:
```
michaeldaas, line 55: i guess this can cause an infinite loop if there are mismatches in termination definitions? maybe better to use a break based on timeout or number of loops here?

michaeldaas (changes requested):
fine for me apart from the whitespace changes and the infinite loop issue.

michaeldaas:
wait, why are there commits from 2 years ago being pushed with this PR?

examples/lab_devices/SignatoneProber.yaml
Marrkson: Any suggestions of which comments/examples might be missing?

Marrkson:
Good point, I think this is from the way I rebased. Has however no effect on todays status of the code.

examples/lab_devices/SignatoneProber.yaml
michaeldaas: you only added the `yaml` file, which dowsn't work as an example on it's own. you could add another (commented out) instantiation to the actual callable example in `probestation.py`, as is done for most other lab devices. it's not that imrportant, though.

michaeldaas (approved):
looks good, thanks for the changes

examples/lab_devices/SignatoneProber.yaml
```diff
@@ -0,0 +1,13 @@
...
+    init     :
+        resource_name : "TCPIP0::169.254.55.86::9090::SOCKET"
+        backend : "@py"
+        read_termination : ""
+        timeout: 2.0
+
+hw_drivers:
+  - name      : Prober
+    type      : SignatoneProber
+    interface : Visa
```
themperek, line 13:
```suggestion
    interface : Visa

```

examples/lab_devices/ProbeStation.py
```diff
@@ -25,3 +25,8 @@
 dut.init()
 print(dut['SussProber'].get_position())
 print(dut['SussProber'].get_die())
+
+dut2 = Dut('SignatoneProber.yaml')
+dut2.init()
+print(dut2['SignatoneProber'].get_position())
+print(dut2['SignatoneProber'].get_die())
```
themperek, line 32:
```suggestion
print(dut2['SignatoneProber'].get_die())

```

DavidLP (approved):
Great!

PR #167, opened 2022-01-18
Support for cocotb-bus 0.2.0

ChristianBesp:
This fix adds compatibility with the latest version of cocotb-bus which is installed when installing cocotb. They introduced case-insensitive signal names on a bus in 0.2.0 which tries to match lowercase names by default.
At the same time it adopts the new assignment sytax, replacing `<=`.

### DIFF
--- a/basil/utils/sim/BasilSbusDriver.py
+++ b/basil/utils/sim/BasilSbusDriver.py
@@ -34,16 +34,16 @@ class BasilSbusDriver(BusDriver):
 
     async def init(self):
         # Defaults
-        self.bus.BUS_RST <= 1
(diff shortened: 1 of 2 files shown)


themperek (approved):

themperek:
MAybe other buses too?

ChristianBesp:
True, there are some more. The SBusDriver is the only one I am using. The same keyword argument `case_insensitive=False` could be added in every place where a BusDriver is initialized. Other option is to go with what cocotb seems to assume as default now (https://github.com/cocotb/cocotb/pull/2138) and switch to lowercase signal names.

Issue #166, opened 2022-01-11
Q: Possibility to add formatter to device description yaml


leloup314:
When using `basil` to read some quantity e.g. current off of an SMU, the resulting return string format depends on the SMU model. This is inconvenient when using different SMU models, even though their API through `basil` is always the same e.g `smu.get_current()`. In my measurement script I end up with something like this:

    if device_type == 'keithley_2410':
        return float(result.split(',')[1])

    elif device_type == 'keithley_6517A':
        return float(result.split(',')[0][-4])
    ...

Is it possible to add something like a `formatter` to the device description yaml from which the the return value of `get_some_quantity` is formatted in such a way that the resulting string contains just the respective quantity's value? e.g.
    
    # Keithley 2410
    smu.get_current()  # Returns '1.23'
    
    # Keithley 6517A
    smu.get_current()  # Returns '1.23'

DavidLP:
I always wanted to add a parser feature that a format string of return values can be defined in a generic way. Since I did not find a suitable python parser package quickly I skipped this endeavor. One could define how to parse the return value in the yaml file

DavidLP:
Maybe there is something like this already there, without defining your own generic parser

leloup314:
Maybe some `regex` would be useful

leloup314:
This has been merged. Closing

PR #168, opened 2022-02-03
Allow optional, device-specific formatting of SCPI query return strings

leloup314:
This PR addresses #166. It enables optional, device-specific formatting of SCPI query return strings for SCPI-type devices which are created from a device description yaml instead of having a dedicated hardware layer Python script. This is helpful if devices return *multiple* fields per query of which only one corresponds to the desired information.

## API
This PR adds the following API for classes inheriting from `basil.HL.scpi.scpi` base class, referred to as `scpi` in the following:

    scpi.has_formatting -> bool : Attribute, whether formatting exists for this SCPI-device
    scpi.formatting_enabled -> bool : Attribute, whether SCPI-query formatting is enabled for this instance
    scpi.enable_formatting() -> None : Method, enables formatting, raises AttributeError if no formatting exists
    scpi.disable_formatting() -> None: Method, disables formatting

## Specifying formatting
Add the following structure to the device descriptor yaml:

    __scpi_query_fmt:
      fmt_sep: ','  # The separator value for multi-value return strings
      fmt_method:  # Define the format string for each method; currently only indexing is supported
        get_voltage: '{0}'  # Voltage information is at index 0 of multi-value string for this device
        get_current: '{1}'  # Voltage information is at index 0 of multi-value string for this device
        ...  # Add more methods who return multi-value strings

Adding formatting is optional.

## Enabling/disabling formatting
Formatting can be enabled after initialization of the SCPI device by calling

    scpi.enable_formatting()
    scpi.disable_formatting()

To automatically enable formatting, pass `enable_formatting: true` to the `init`-field in the  device config yamls `hw_driver`:

    # Snippet
    hw_drivers:
      - name      : Sourcemeter
        type      : scpi
        interface : Visa
        init      :
            device : Keithley 2410
            enable_formatting: true

## Example
A Keithley 2410 SMU  is generated from [this](https://github.com/SiLab-Bonn/basil/blob/master/basil/HL/keithley_2410.yaml) device descriptor yaml which contains formatting. Reading a voltage from the SMU with basil (using [this](https://github.com/SiLab-Bonn/basil/blob/master/examples/lab_devices/keithley2410_pyvisa.yaml) `smu_config`) looks like this:

    # ... do the imports and define smu_config

    # Create device
    smu = Dut(smu_config)
    smu.init()
    
    # Read voltage
    smu['Sourcemeter'].get_voltage()
    
    # Prints something like '-5.124E-05,+1.789E-10,+1.001E37,1.001E5'

The desired information, the voltage, is found in the 0th field of the return string. Consequently, the following code is often found with basil

    voltage = float(smu['Sourcemeter'].get_voltage().split(',')[0])

This PR enables to define specific formatting per method in the [device descriptor yaml of the e.g. Keithley 2410](https://github.com/leloup314/basil/blob/master/basil/HL/keithley_2410.yaml#L30) so that the call changes to

    # Enable formatting
    smu['sourcemeter'].enable_formatting()

    # Read voltage
    voltage = smu['Sourcemeter'].get_voltage()  # Returns '-5.124E-05'

    # Convert to float
    voltage = float(voltage)

which corresponds to a more consistent behavior with respect to the method called.

### DIFF
--- a/tests/test_SimSCPIFormatting.py
+++ b/tests/test_SimSCPIFormatting.py
@@ -25,7 +25,7 @@ hw_drivers:
 '''
 
 
-class TestSimScpiFormatting(unittest.TestCase):
+class TestSimScpi(unittest.TestCase):
 
     def setUp(self):
         self.cfg = yaml.safe_load(cnfg_yaml)
--- a/tests/test_SimSCPIFormatting.py
+++ b/tests/test_SimSCPIFormatting.py
@@ -47,7 +47,7 @@ class TestSimScpi(unittest.TestCase):
         # Disable formatting
         self.device['Sourcemeter'].disable_formatting()
         self.assertEqual(voltage, '-5.124E-05')
-    
+
 
 if __name__ == '__main__':
     unittest.main()
--- a/tests/test_SimSCPIFormatting.py
+++ b/tests/test_SimSCPIFormatting.py
@@ -21,32 +21,43 @@ hw_drivers:
     interface : Visa
     init      :
         device : Keithley 2410
+        device_formatting: true
 '''
 
 
-class TestSimScpi(unittest.TestCase):
+class TestSimScpiFormatting(unittest.TestCase):
 
     def setUp(self):
         self.cfg = yaml.safe_load(cnfg_yaml)
         self.device = Dut(self.cfg)
         self.device.init()
+        # Check that formatting is present
+        self.assertTrue(self.device['Sourcemeter'].has_formatting)
+        # Check that formatting is enabled after init
+        self.assertTrue(self.device['Sourcemeter'].formatting_enabled)
 
     def tearDown(self):
         self.device.close()
 
     def test_read_voltage(self):
-        voltage = self.device['Sourcemeter'].get_voltage().split(',')[0]
+        voltage = self.device['Sourcemeter'].get_voltage()
         self.assertEqual(voltage, '-5.124E-05')
 
-    def test_read_voltage_formatted(self):
-        # Check that formatting is present
-        self.assertTrue(self.device['Sourcemeter'].has_formatting)
-        # Enable formatting
-        self.device['Sourcemeter'].enable_formatting()
-        voltage = self.device['Sourcemeter'].get_voltage()
+    def test_read_current(self):
+        voltage = self.device['Sourcemeter'].get_current()
+        self.assertEqual(voltage, '+1.789E-10')
+
+    def test_read_voltage_unformatted(self):
+        # Check that formatting is enabled
+        self.assertTrue(self.device['Sourcemeter'].formatting_enabled)
         # Disable formatting
         self.device['Sourcemeter'].disable_formatting()
+        voltage = self.device['Sourcemeter'].get_voltage().split(',')[0]
         self.assertEqual(voltage, '-5.124E-05')
+        # Check that formatting is disabled
+        self.assertFalse(self.device['Sourcemeter'].formatting_enabled)
+        # Enable formatting
+        self.device['Sourcemeter'].enable_formatting()
 
 
 if __name__ == '__main__':
--- a/tests/test_SimSCPIFormatting.py
+++ b/tests/test_SimSCPIFormatting.py
@@ -43,10 +43,6 @@ class TestSimScpi(unittest.TestCase):
         voltage = self.device['Sourcemeter'].get_voltage()
         self.assertEqual(voltage, '-5.124E-05')
 
-    def test_read_current(self):
-        voltage = self.device['Sourcemeter'].get_current()
-        self.assertEqual(voltage, '+1.789E-10')
-
     def test_read_voltage_unformatted(self):
         # Check that formatting is enabled
         self.assertTrue(self.device['Sourcemeter'].formatting_enabled)
--- a/tests/test_SimSCPIFormatting.py
+++ b/tests/test_SimSCPIFormatting.py
@@ -21,7 +21,7 @@ hw_drivers:
     interface : Visa
     init      :
         device : Keithley 2410
-        device_formatting: true
+        enable_formatting: true
 '''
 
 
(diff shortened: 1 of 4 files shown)


leloup314:
What do you think @themperek ?

themperek:
This is was newer taken care of by me but is as-is. Looks OK to me. 
It will break backwards compatibility?
Can we have some tests for this?

leloup314:
It is backwards compatible with Python 2.7. Since it doesn't change the API in any way but only alters the return string of a SCPI query I find it hard to think of a test for this. It is to be defined in the device descriptors inside basil and is optional. What it will break, is code where people call `device['Sourcemeter'].get_current().split(',')[1]` but this is fine IMHO because the behaviour is inconsistent: The user must know the return format of e.g. Keithley 2410 to know which one is the current even though the person called `get_current` on the smu

themperek:
>It is backwards compatible with Python 2.7. 

I mean backwards compatible for peoples code as they have now.
Exactly this `What it will break, is code where people call device['Sourcemeter'].get_current().split(',')[1]`

For me this is fine but maybe users should say?

themperek:
This is not safe but maybe one can include routine in `yaml` and call `eval()` on this? Would be versatile.

leloup314:
I made it optional now so it is backwards compatible and will not break any existing implementation. The following methods /attributes are implemented

- `has_formatting` -> bool : Whether or not formatting is specified for this SCPI device in the description
- `enable_formatting` -> None: Enables formatting, raises `AttributeError` if `has_formatter` is `False`  
- `disable_formatting` -> None: Disables formatting

leloup314:
I would strongly argue against use of `eval()`  for safety reasons. I also think it would be overkill because what one really only need is

1. Separate a multi value string by the SCPI-query-specific separator token
2. Extract the relevant data at given index

themperek:
>it doesn't change the API in any way but only alters the return string of a SCPI query I find it hard to think of a test for this.

Can we have an example for this?

In general, would be great to think about how one could test those parts? @cbespin @YannickDieter @idcs0906

cbespin:
> In general, would be great to think about how one could test those parts?
The first idea that comes to my mind would be mocking these devices, either on device or transfer layer. One needs to check for all supported devices various return strings for serial commands, though.
Easier is if we start at return string level and just test the formatting, but would need to know the return strings.

themperek:
> > In general, would be great to think about how one could test those parts?
> 
> The first idea that comes to my mind would be mocking these devices, either on device or transfer layer. One needs to check for all supported devices various return strings for serial commands, though. Easier is if we start at return string level and just test the formatting, but would need to know the return strings.

It does not have to be any fancy at all.
We have this https://github.com/SiLab-Bonn/basil/blob/master/tests/test_SimScpi.py
And if I am not wrong can easily build your own devices? https://github.com/SiLab-Bonn/basil/blob/master/tests/test_SimScpi.py

leloup314:
> Easier is if we start at return string level and just test the formatting, but would need to know the return strings.

I think this is what we should do. We really can only test easily that the implementation of optional formatting works for a known return string and specified formatting. Formatting is specified per device as well as per method, generated for the respective SCPI query. To test this you would need to know the return string for each method for which formatting is defined.

themperek:
This seems to work:

**test_sim_keithlay2410.py:**
```python3

import unittest
import yaml
import os
from basil.dut import Dut

k2410def_yaml = os.path.join(os.path.dirname(__file__), "k2410.yaml")

cnfg_yaml = f'''
transfer_layer:
  - name     : Visa
    type     : Visa
    init     :
        resource_name : ASRL1::INSTR
        read_termination : "\\n"
        write_termination : "\\r\\n"
        backend : "{k2410def_yaml}@sim"

hw_drivers:
  - name      : Sourcemeter
    type      : scpi
    interface : Visa
    init      :
        device : Keithley 2410
'''

class TestSimScpi(unittest.TestCase):

    def setUp(self):
        self.cfg = yaml.safe_load(cnfg_yaml)
        self.device = Dut(self.cfg)
        self.device.init()

    def tearDown(self):
        self.device.close()

    def test_read_voltage(self):
        voltage = self.device['Sourcemeter'].get_voltage().split(',')[0]
        self.assertEqual(voltage, '-5.124E-05')

if __name__ == '__main__':
    unittest.main()
```
**k2410.yaml:**
```yaml
spec: "1.1"
devices:
  device 1:
    eom:
      ASRL INSTR:
        q: "\r\n"
        r: "\n"
      USB INSTR:
        q: "\n"
        r: "\n"
      TCPIP INSTR:
        q: "\n"
        r: "\n"
      GPIB INSTR:
        q: "\n"
        r: "\n"
    dialogues:
      - q: "*IDN?"
        r: "KEITHLEY INSTRUMENTS INC.,MODEL 2410"
    properties:
      selected_channel:
        default: 'VOLT'
        setter:
          q: "SENSE:FUNC '{}'"
    channels:
      type1:
        ids: ['VOLT']
        can_select: False
        properties:
          value:
            default: 1.0
            getter:
              q: ':READ?'
              r: '-5.124E-05,+1.789E-10,+1.001E37,1.001E5'

resources:
  ASRL1::INSTR:
    device: device 1
```

tests/test_SimSCPIFormatting.py
```diff
@@ -0,0 +1,53 @@
...
+        read_termination : "\\n"
+        write_termination : "\\r\\n"
+        backend : "{k2410def_yaml}@sim"
+
+hw_drivers:
+  - name      : Sourcemeter
+    type      : scpi
+    interface : Visa
+    init      :
+        device : Keithley 2410
```
themperek, line 23: Would there be a way to automatically enable formating by setting something here?

themperek (approved):

tests/test_SimSCPIFormatting.py
leloup314: One could enable it from the device configuration, yes. Maybe by defining something like `enable_formatting: true`

themperek: I think it is OK to go as is. You can add this `enable_formatting: true` for your own convenience.

themperek:
Looks great. Is this still WIP?

leloup314:
You can merge. I was thinking of adding formatting for as many devices as possible now, but I will do this separately

PR #169, opened 2022-02-03
Add HP 4284A LCR meter to basil

leloup314:
This PR add the Hewlett-Packard 44284A precision LCR meter to basil. Unfortunately, it currently only works relatively straight forward using GPIB driver under Windows.


leloup314:
Tested today, LCR meter works with basil now

leloup314:
@themperek you can merge, I even cleaned up my git ;)

PR #171, opened 2022-02-09
FIX: remove NTC temperature logging

### LINKED ISSUE #170: Unnecessary NTC Register temperature logging
The [NTCRegister.py](https://github.com/SiLab-Bonn/basil/blob/master/basil/HL/NTCRegister.py) module spams the terminal with logging of the temperature when calling `get_temperature()`, since one usually calls the function a number of times and computes the average. The function should probably return the temperature and leave logging to the user.

lschall:
Remove unnecessary logging in `get_temerpature` function of NTCRegister due to spamming of consol. Logging should be done by user.

Closes #170


themperek:
@lschall  First commit 🎉

PR #174, opened 2022-02-17
ADD: lab devices example for Keithley 2410 with formatting

leloup314:
This  PR adds `enable_formatting: true` to the existing example device config for the Keithley 2410 SMU. Furthermore, it adds an example device config for Keithley 2410 for PySerial TF.


leloup314:
@themperek are you doing a rebase and merge on a PR?

PR #178, opened 2022-06-01
Keithley 6517a electrometer

SinuoZhang:
The pull request contains the changes of 3 files which are related to the keithley6517a electrometer.
It includes:
   basil/HL/keithley_6517a.yaml : the commands with some comments
   examples/lab_devices/keithley6517a_pyvisa.yaml : an example for setting up the remote connection
   examples/lab_devices/keithley6517a.py : an example for the application


SinuoZhang:
examples are removed from the HL yaml. The equivalent examples are in the example python file

examples/lab_devices/keithley6517a_pyvisa.yaml
```diff
@@ -0,0 +1,22 @@
...
+transfer_layer:
+  - name: electrometer
+    type: Visa
+    init:
+        # using serial-USB adapter
+        resource_name: ASRLCOM6         # example for windows system
+#        resource_name: ASRL/dev/ttyUSB0 # example for linux system 
+        timeout : 1000
+        read_termination: "\n"
+        baud_rate: 19200
```
leloup314, line 15: Is 19200 the standard baud rate of the device when powered on or did you set this?

examples/lab_devices/keithley6517a.py
```diff
@@ -0,0 +1,76 @@
...
+
+time.sleep(0.3)
+dut['EMeter'].select_current()          # measure current
+dut['EMeter'].connect_meter()           # connect the source and meter, resulting in a simpler circuit
+dut['EMeter'].set_source_range('MAX')   # set the output limit of the voltage source to 1000V (MAX)
+
+# Perform zero check
+# set the smallest range for zero check
+dut['EMeter'].set_current_range(20e-12)     # unit is A (everywhere)
+time.sleep(0.5)
```
leloup314, line 50: Does the manual say anything about waiting times for certain operations?

leloup314: Can you check something like an `is_ready` parameter on the device instead?

examples/lab_devices/keithley6517a_pyvisa.yaml
SinuoZhang: is the default. the read termination cannot be set, so the value there is based on experience

examples/lab_devices/keithley6517a.py
SinuoZhang: I didn't notice about a standard setting of the waiting time. So these are empirical, set the time until it stops complaining. maybe at some points I set a slightly too long time, but will work

examples/lab_devices/keithley6517a.py
```diff
@@ -0,0 +1,76 @@
+###################################################
+# This is an example for keithley6517a electrometer
+# Last Modified: Mi 01 Jun 2022 07:30:23  CEST
+###################################################
+
+import time
+from basil.dut import Dut
+from basil.TL.Visa import visa
```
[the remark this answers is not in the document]
SinuoZhang: yea, seems the case. I remove this

examples/lab_devices/keithley6517a.py
SinuoZhang: > Can you check something like an `is_ready` parameter on the device instead?

Is there any device using this feature, it would be faster to check what command it uses and then look for the similar commands, as long as it's SCPI

leloup314: There is something in SCPI maybe you can check the two options mentioned [here](https://stackoverflow.com/questions/30725746/is-it-normal-to-wait-for-some-time-between-operations-under-the-visa-protocol)

leloup314: @s6thseng maybe the solution in the link can help you to sync your R&S PSU so you don't have to put arbitrary sleeps and/or try many times

SinuoZhang: The suggested SCPI commands are "*WAI" and "*OPC".
I've tried both commands and they doesn't seems working. According to the manual, these 2 commands are especially useful for the so called "overlapped commands":
Screenshot 2022-06-02 at 13 48 57
The test I've done purely consists of the other type (sequential commands), which should not need the "waiting" commands according to the manual. As a result of the test, error messages showed up as if no "waiting" commands were applied.
I've also tested the possibility of wrong command format, by adding an additional time sleep. No error message has appeared.
I guess there are probably a smarter way to do this "waiting" procedure, but it seems less obvious.
For now, I would say that the time sleep could be a good alternative, since the setting only needs to be made once in the script.

BY THE WAY, please don't merge them now, I found some mistakes in the example, I will fix them first

s6thseng: Tried your proposed solution, but still does not work. To sleep is still the easiest solution.

examples/lab_devices/keithley6517a_pyvisa.yaml
leloup314: Is the `timeout` in ms?

examples/lab_devices/keithley6517a.py
leloup314: For VISA you can set a `query_delay` in the transfer layer config's `init`, can you test if this removes the need for manual sleeps?

examples/lab_devices/keithley6517a_pyvisa.yaml
leloup314: Can you add a `query_delay: 1.0` to the transfer layer `init`? This should do the sleeping you are putting manually

leloup314:
Let me know once this is ready to merge. Please also check why the [codestyle test](https://github.com/SiLab-Bonn/basil/runs/6709941923?check_suite_focus=true) is failing and fix it ;)

examples/lab_devices/keithley6517a.py
s6thseng: seems not to work for the PS

SinuoZhang:
@leloup314 The code style error were fixed. The example script was tested. It should be okay to merge, if you don't have other remarks.

leloup314:
I will merge this later! Thank you for contributing!

PR #180, opened 2022-07-06
ENH: better API for NTCReadout allowing to get/set all needed quantit…

leloup314:
…ies in firmware


leloup314:
@s6thseng testing is required ;) Just check if the example runs after flashing the new firmware

themperek (approved):

PR #181, opened 2022-08-31
Tektronix simplify

Marrkson:
This PR simplifies support for all tektroniks oscilloscopes in Silab and enables easy retrieval of raw curve data from the oscilloscope, to generate plots later on the PC.


basil/HL/tektronix_oscilloscope.py
```diff
@@ -0,0 +1,197 @@
...
+
+    def _get_xscale(self):
+        """Get the horizontal scale of the instrument"""
+
+        # This function will generate a VISA timeout if there is no waveform data
+        # available and the channel is enabled.
+        result = None
+        try:
+            xincr = self._intf.query("WFMOutpre:XINCR?").strip()
+        except Exception:
```
leloup314, line 98: No bare `except` statement. Please catch the timeout generated by pyvisa, otherwise hard to debug if things are not working

basil/HL/tektronix_oscilloscope.py
```diff
@@ -0,0 +1,197 @@
...
+        This function should be called before performing a curve query."""
+        # Use "display:global:CH{x}:state?" to determine if the channel is displayed
+        # and available for download
+        if source == "NONE":
+            display_on = False
+        else:
+            display_on = bool(
+                int(self._intf.query("display:global:{}:state?".format(source)))
+            )
+        return display_on
```
leloup314, line 131: A more pythonic solution to implement this would be to return immediately, like:

    def has_data_available(self, source):
        if source == 'None':
            return False
        return bool(int(self._intf.query("display:global:{}:state?".format(source))))

basil/HL/tektronix_oscilloscope.py
```diff
@@ -0,0 +1,197 @@
...
+                )
+
+            else:
+                raise Exception(
+                    "It should have been impossible to execute this code"
+                )
+
+        # # Restore the acquisition state
+        self._intf.write("ACQuire:STATE {}".format(acq_state))
+        return ret_val
```
leloup314, line 197: This function can be simplified by

    ....
    
    if wave_type is WaveType.ANALOG:
        ret_val = self._post_process_analog(source_data, x_scale, channel=channel)
    else:
        raise NotImplementedError(f"Analysis for type {wave_type} data not yet implemented!")
    
    # Restore the acquisition state
    self._intf.write("ACQuire:STATE {}".format(acq_state))
    return ret_val

Also, I would never write an Exception for code which should not be executed in production code, this is more like a debug strategy ;)

leloup314 (changes requested):

basil/HL/tektronix_oscilloscope.py
```diff
@@ -0,0 +1,183 @@
...
+        jobs = self._make_jobs(channel=channel)
+
+        # remember the state of the acquisition system and then stop acquiring waveforms
+        acq_state = self._intf.query("ACQuire:STATE?").strip()
+        self._intf.write("ACQuire:STATE STOP")
+        wave_type = self._setup_curve_query(jobs, channel=channel)
+
+        # Horizontal scale information
+        x_scale = self._get_xscale()
+        if x_scale is not None:
```
leloup314, line 174: In case that `x_scale` is None, trying to return `ret_val` will result in a `NameError`. Either declare it in the beginning as `None` or move the return statement inside the `if x_scale is not None` branch

leloup314 (changes requested):

leloup314 (approved):

leloup314:
Has this been tested on the actual machine?

PR #182, opened 2022-08-31
Binder mk56

Marrkson:
ADD basil support for the Binder MK56 in the basement, based on the exisitng MK53

### DIFF
--- a/basil/HL/binder_mk53.py
+++ b/basil/HL/binder_mk53.py
@@ -15,14 +15,6 @@ FUNCTION_READN_ALT = 0x04  # read n words, does not occur
 FUNCTION_WRITE = 0x06  # write word
 FUNCTION_WRITEN = 0x10  # write n words
 
-# Operation addresses
-ADDR_CURTEMP = 0x11a9
-ADDR_DOOROPEN = 0x1007  # does not work, wrong address?
-ADDR_SETPOINT = 0x1077
-ADDR_MANSETPT = 0x1581
-ADDR_BASICSETPT = 0x156f
-ADDR_MODE = 0x1a22
-
 # Error codes
 ERROR_CODES = {
     1: "Invalid function",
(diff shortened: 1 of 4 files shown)


basil/HL/binder_mk53.py
```diff
@@ -45,21 +37,36 @@ def __init__(self, intf, conf):
 
     def init(self):
         super(binderMK53, self).init()
+        # Operation addresses
+        self.ADDR_CURTEMP = 0x11a9
+        self.ADDR_DOOROPEN = 0x1007  # does not work, wrong address?
+        self.ADDR_SETPOINT = 0x1077
+        self.ADDR_MANSETPT = 0x1581
+        self.ADDR_BASICSETPT = 0x156f
+        self.ADDR_MODE = 0x1a22
```
leloup314, line 46: Why did you move the `ADDR` attributes into the instance but not the `FUNCTION` ones and `ERROR_CODES`? If you want to be consistent I also would put these attributes in the class (since they are the same for every instance) like so

    class BinderMK53(HardwareLayer):

        # MODBus function codes
        FUNCTION_READN = 0x03  # read n words
        FUNCTION_READN_ALT = 0x04  # read n words, does not occur
        FUNCTION_WRITE = 0x06  # write word
        FUNCTION_WRITEN = 0x10  # write n words

        # Operation addresses
        ADDR_CURTEMP = 0x11a9
        ADDR_DOOROPEN = 0x1007  # does not work, wrong address?
        ADDR_SETPOINT = 0x1077
        ADDR_MANSETPT = 0x1581
        ADDR_BASICSETPT = 0x156f
        ADDR_MODE = 0x1a22

        # Error codes
        ERROR_CODES = {
            1: "Invalid function",
            2: "Invalid parameter address",
            3: "Pparameter value outside range of values",
            4: "Slave not ready",
            5: "Write access to parameter denied"
        }

        def init(self):
            super(binderMK53, self).init()

Now all off these are avbailable in each instance via `self.ADDR*`, `self.FUNCTION*` and `self.ERROR_CODES`.
The advantage is now that you do not have to define everything again for your BinderMK56 (which is a subclass of BinderMK53) since almost everything, except 3 `ADDR` attributes, is the same

basil/HL/binder_mk56.py
```diff
@@ -0,0 +1,110 @@
...
+FUNCTION_READN = 0x03  # read n words
+FUNCTION_READN_ALT = 0x04  # read n words, does not occur
+FUNCTION_WRITE = 0x06  # write word
+FUNCTION_WRITEN = 0x10  # write n words
+
+# Error codes
+ERROR_CODES = {
+    1: "Invalid function",
+    2: "Invalid parameter address",
+    3: "Pparameter value outside range of values",
```
leloup314, line 22: Typo

basil/HL/binder_mk56.py
```diff
@@ -0,0 +1,110 @@
...
+        if temperature < self.min_temp:
+            raise RuntimeWarning('Set temperature %f is lower than minimum allowed temperature %f' % (temperature, self.min_temp))
+        if temperature > self.max_temp:
+            raise RuntimeWarning('Set temperature %f is higher than maximum allowed temperature %f' % (temperature, self.max_temp))
+        for _ in range(reps):
+            try:
+                self.write(self.ADDR_MANSETPT, self._encode_float(temperature))
+                break
+            except RuntimeWarning:
+                pass
```
leloup314, line 64: The exceptions raised here should be ValueErrors not RuntimeWarnings. If you only want to warn and not crash I suggest you use the `warnings` module.

leloup314: Also the `%` style formatting is deprecated, nowadays use f-strings or the format-Method

basil/TL/Socket.py
```diff
@@ -41,13 +42,18 @@ def write(self, data):
...
     def read(self, buffer_size=1):
         ret = self._sock.recv(buffer_size)
-        data = ret.split(self.read_termination.encode(self.encoding))
-        return data[:-1]
+        if not self.handle_as_byte:
+            data = ret.split(self.read_termination.encode(self.encoding))
+            data = data[:-1]
+        else:
+            data = ret
+        return data
```
leloup314, line 56: Simplify to

    ....

    if not self.handle_as_byte:
        return ret.split(self.read_termination.encode(self.encoding))[:-1]
    return ret

leloup314 (changes requested):

basil/HL/binder_mk53.py
Marrkson: Done.

basil/TL/Socket.py
Marrkson: Done.

leloup314 (approved):

leloup314:
Has the latest commit been tested on the actual machine?

PR #189, opened 2022-11-30
new julabo fp50

AntonioT7:
communicate with chiller in the measurement room


leloup314:
@AntonioT7 maybe you can fix the flake8? ;)

leloup314:
Nice, thank you for fixing the flake8. I will not merge yet because there are still some things which we should talk about in person. As a general remark, you should always state what has been done in the commit message instead of just repeating it for every push. How we usually write commits in SiLab is explained [here](https://github.com/SiLab-Bonn/pyBAR/wiki#policy-for-commit-messages)

PR #190, opened 2023-01-13
Add cross point switch HL for EOS and QMS card

matthias-schuessler:
Adds a file XPT.py containing the HL for EOS and QMS card.

Also contains a change to Keithley2000 adding the option to set NPLC in DC voltage mode

### DIFF
--- a/basil/HL/XPT.py
+++ b/basil/HL/XPT.py
@@ -9,13 +9,15 @@
 import yaml
 from array import array
 from basil.HL.HardwareLayer import HardwareLayer
-from struct import pack, unpack, unpack_from, calcsize
+from struct import pack, unpack_from
 import logging
+
 
 class I2C_INTF(HardwareLayer):
     '''
     Use basil implementation of i2c.
     '''
+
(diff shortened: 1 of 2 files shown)


leloup314:
Thanks for the contribution,  looks good! Would be nice if you could add an example to the [examples folder](https://github.com/SiLab-Bonn/basil/tree/master/examples). If you don't find any of the existing folders fitting (e.g. `lab_devices`, etc) you can also add a new folder (e.g. `EOS` or `serial_powering`, etc). The example should consist of a config-yaml as well as a Python script showcasing the usage and basic functionality of the contents of the `XPT.py` HL.

matthias-schuessler:
I talked with Florian and Yannick about adding an example. We think that it doesn't make sense to add one, since our only application for the switches is the QMS-card, which needs bdaq to operate. Thus, an example wouldn't work in basil without adding the entire bdaq class.

leloup314:
Okay, then let me know once this is ready to be merged

matthias-schuessler:
From my end, this is ready to merge.

PR #191, opened 2023-02-06
Agilent33250a.py fix

dschuechter:
Fixing a small leftover from python 2 times. In python 3 `string.atof()` is not supported anymore and can be replaced by `float()`.


leloup314:
Thanks for the PR! As you can see, `flake8` is failing because now the imported `string` is unused. Could you remove the import? This should fix the `flake8` job

dschuechter:
Updated!

leloup314:
Very nice, thank you for contributing!

PR #193, opened 2023-02-21
agilent_33250a.yaml addded missing commands

dschuechter:
Updated agilent_33250a.yaml to be able to toggle the output within basil


leloup314:
Cool, thanks! If I remember correctly, there was also an issue using the example [description yaml for the oszilloscope](https://github.com/SiLab-Bonn/basil/blob/master/examples/lab_devices/tektronixOscilloscope_pyvisa.yaml), right? Could you maybe also adapt this within this PR?

dschuechter:
I can briefly create an example for ```examples/lab_devices```. I had to create a new ```tektronix_tds3034b.py``` in ```basil/HL/``` since the older oscilloscope returns values differently.

So in total there would be the following new files:
```examples/lab_devices/tektronix_tds3034B.yaml```
```examples/lab_devices/tektronix_oscilloscope_tds3034b.py```
```basil/HL/tektronix_tds3034B.yaml```

I have not tested all commands. Some functions of the newer oscilloscopes might not work. DAQ and controlling standard settings like Scale, Pos, CH, Trigger,... work.

I thought that this should be a new PR

leloup314:
In that case you are right, it should be a different PR! Thanks!

PR #194, opened 2023-02-21
PRJ: Added support for Tektronix TDS3034B

dschuechter:
Added support for the Tektronix TDS3034B oscilloscope incl. an example.

It shares the same device ```.yaml``` file as the newer Tektronix oscilloscopes. 
For that reason, some commands that work on newer devices might not be compatible with the old TDS3034B. 
Standard tasks like DAQ, configuration of the oscilloscope and so on are working.


leloup314:
Thanks! Can you have a look at the `flake8` ?

dschuechter:
```flake8``` is now fixed.

I will look after ```flake8``` in future PR's :sweat_smile:

leloup314:
Thank you! Looks good!

PR #202, opened 2023-04-25
Improve exception handling for dynamic method generation

leloup314:
This PR adds more streamlined exception handling for the SCPI `HL`.

Currently, when the dynamically-generated method fails to create a SCPI command because of **any** exception, the bare exception is excepted and a `ValueError` is raised within the exception handling. Due to Pythons [exception chaning](https://docs.python.org/3.1/whatsnew/3.0.html#changes-to-exceptions), this produces a confusing traceback, not clearlyindictaing that _no such attribute is available_.

This PR changes this behavior by 

1. Catching and handling the possible exceptions occurring on SCPI command creation
2. Checking for successful SCPI command creation and  raising an `AttributeError` if no such SCPI method is available

This mirrors the expected behaviour of `arbitrary_object.non_existing_method()` generating an `AttributeError`

### DIFF
--- a/basil/HL/scpi.py
+++ b/basil/HL/scpi.py
@@ -75,10 +75,17 @@ class scpi(HardwareLayer):
         '''
         def method(*args, **kwargs):
             channel = kwargs.pop('channel', None)
+            command = None
             try:
-                command = self._scpi_commands['channel %s' % channel][name] if channel is not None else self._scpi_commands[name]
-            except Exception:
-                raise ValueError('SCPI command %s is not defined for device %s' % (name, self.name))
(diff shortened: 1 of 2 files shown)


leloup314:
@themperek any reason that `SCPI` hardware layer shout not raise an `AttributeError` when dynamically-generated method from `__getattr__` tries to call SCPI command that does not exist?

themperek:
@leloup314  I would be fine with this but this part is from @DavidLP

leloup314:
Probably will break peoples code... I guess it is not worth to change it for now.

PR #205, opened 2023-11-27
Create branch for backward compatibility of the MIO1.04 board

dschuechter:
Some firmware modules were backported to be compatible with the Spartan 3 FPGA on MIO1.04


leloup314:
Please fix the conflict in your fork, then we can merge

PR #209, opened 2024-01-09
serial and visa baud rate init parameter compatibility

mjmucha:
Currently the baud rate init parameter is set for the serial transfer layer using `baudrate` and for the visa transfer layer using `baud_rate`.

This pull request makes these two variants interchangeable for the serial and the visa transfer layer with the aim to improve the consistency of the interface

### DIFF
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.8]
+        python-version: [3.11]
     steps:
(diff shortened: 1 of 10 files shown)


YannickDieter:

cbespin (approved):
Looks good to me. We should investigate and decide if we want a default `baud_rate`/`baudrate` key for all classes and internally rename for the respective library in the future. If I see it correctly, `pyvisa` and `pyserial` use `baud_rate` and `baudrate`, respectively. At the moment (with this PR), we just allow the user to use them interchangeably.

Issue #188, opened 2022-11-08
Pyserial module not listed in the requirements


Giakou:
Hi, is there any reason why `pyserial` is not listed in the _requirements.txt_? I have never used basil before, but I just cloned it in development mode and tried to execute a copy of `basil/examples/lab_devices/xray_tube.py`. This is the traceback:
```
Traceback (most recent call last):
  File "/home/pxd9/basil/basil/dut.py", line 255, in _factory
    mod = import_module(importname)
  File "/home/pxd9/anaconda3/envs/mosfet_irrad/lib/python3.10/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1006, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 688, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 883, in exec_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "/home/pxd9/basil/basil/TL/Serial.py", line 9, in <module>
    import serial
ModuleNotFoundError: No module named 'serial'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/pxd9/mosfet_irrad/irrad/xray_irrad.py", line 35, in <module>
    device = Dut('xray_basil_config.yaml')
  File "/home/pxd9/basil/basil/dut.py", line 101, in __init__
    self.load_hw_configuration(self._conf)
  File "/home/pxd9/basil/basil/dut.py", line 200, in load_hw_configuration
    self._transfer_layer[intf['name']] = self._factory('basil.TL.' + intf['type'], *(), **kargs)
  File "/home/pxd9/basil/basil/dut.py", line 259, in _factory
    mod = import_module(importname)
  File "/home/pxd9/anaconda3/envs/mosfet_irrad/lib/python3.10/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1004, in _find_and_load_unlocked
[...]

PR #210, opened 2024-01-15
ADD: Automatic finding of USB binds based on instrument IDN

mjmucha:
Implemented `findUSBBinds()` to find the Linux USB binds based on given IDN in the configuration file. With `modify_basil_config()` one can then automatically update the ports/resource_names of the transfer layers (serial/visa).

For this to work the transfer layer needs an additional init parameter (`identification`). Like in this example, where `resource_name` is automatically updated to the current USB bind after executing `modify_basil_config()`.

```
transfer_layer:
  - name     : HMP4040
    type     : Visa
    init     :
        resource_name : ASRL/dev/ttyUSB0::INSTR
        read_termination : "\r\n"
        identification: HAMEG,HMP4040,105820,HW50020001/SW2.51
        baud_rate : 38400
        backend : "@py"
```

### DIFF
--- a/basil/utils/USBBinds.py
+++ b/basil/utils/USBBinds.py
@@ -9,7 +9,7 @@ import pyvisa
 import ruamel.yaml
 
 
-def queryIdentification(rm, resource, baud_rate, read_termination="\n", write_termination="\n", timeout=1000 * 5):
+def query_identification(rm, resource, baud_rate, read_termination="\n", write_termination="\n", timeout=1000 * 5):
     """
     Queries the identification of the instrument connected via USB.
 
@@ -34,7 +34,7 @@ def queryIdentification(rm, resource, baud_rate, read_termination="\n", write_te
     return inst.query("*IDN?", delay=0.5)
 
 
-def findUSBBinds(rm, log,
+def find_usb_binds(rm, log,
                  instruments,
                  binds_to_skip=[],
                  memorized_binds=[],
@@ -120,7 +120,7 @@ def findUSBBinds(rm, log,
     return results
 
 
-def getBaudrate(dictionary):
+def get_baudrate(dictionary):
     """
     Gets the baud rate from the given dictionary.
 
--- a/basil/utils/USBBinds.py
+++ b/basil/utils/USBBinds.py
@@ -141,7 +141,7 @@ def modify_basil_config(conf, log, skip_binds=[], verbose=False):
 
         instrument = tf["init"]["identification"]
         baud_rate = getBaudrate(tf["init"])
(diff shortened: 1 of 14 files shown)


YannickDieter:

requirements.txt
```diff
@@ -2,3 +2,5 @@ numpy
 pyyaml
 bitarray>=2.0.0
 six
+pyvisa
```
cbespin, line 5: Why did we not have this before @YannickDieter ? Was there a good reason?

basil/utils/USBBinds.py
```diff
@@ -0,0 +1,245 @@
...
+from pathlib import Path
+
+from six import string_types
+
+import pyvisa
+
+import ruamel.yaml
+
+
+def queryIdentification(rm, resource, baud_rate, read_termination="\n", write_termination="\n", timeout=1000 * 5):
```
cbespin, line 12: Please use pythonic function names (snake case instead of camel case). Also: the default read/write termination defined in the Serial connection driver (https://github.com/SiLab-Bonn/basil/blob/master/basil/TL/Serial.py#L31-L32) is `None`

basil/utils/USBBinds.py
```diff
@@ -0,0 +1,245 @@
...
+    inst = rm.open_resource(resource)
+    inst.timeout = timeout
+    inst.baud_rate = baud_rate
+    inst.read_termination = read_termination
+    inst.write_termination = write_termination
+
+    return inst.query("*IDN?", delay=0.5)
+
+
+def findUSBBinds(rm, log,
```
cbespin, line 37: See above

basil/utils/USBBinds.py
```diff
@@ -0,0 +1,245 @@
...
+                    if len(results) == len(instruments):
+                        return results
+
+            except pyvisa.VisaIOError:
+                pass
+
+    return results
+
+
+def getBaudrate(dictionary):
```
cbespin, line 123: See above

cbespin (changes requested):
This feature will be very useful! I like to keep the amount of required packages as low as possible, but I see no alternative to ruamel.yaml here.
When the function names are changed to something more pythonic, I will approve

requirements.txt
leloup314: Same reason as #188 I would say.

cbespin: I would propose to add the requirements in this way, so that the install contains everything that is required for usage

cbespin: Let's move the discussion to #188. And @mjmucha Can you exclude pyvisa from the `requirements.txt` then? At last for now this keeps consistency.

cbespin (approved):
Looks good now, thanks! The discussion about required packages will be moved to #188

PR #213, opened 2024-04-02
Improvements for mercury driver

SilasM2001:


YannickDieter:

basil/HL/mercury.py
```diff
@@ -56,8 +73,42 @@ def get_channel(self, address=None):
         self._write_command("TS", address)
         return self.read()
 
-    def set_position(self, value, address=None):
+    def set_position(self, value, precision=500, address=None, wait=True):
         self._write_command("MA%d" % value, address)
+        if wait is True:
+            pos = self._wait_FE(address)
+            if abs(pos - value) <= precision:
+                print("At position", pos, "Traget at", value)
```
YannickDieter, line 81: Super picky, but here is a typo ;)

basil/HL/mercury.py
```diff
@@ -56,8 +73,42 @@ def get_channel(self, address=None):
         self._write_command("TS", address)
         return self.read()
 
-    def set_position(self, value, address=None):
+    def set_position(self, value, precision=500, address=None, wait=True):
         self._write_command("MA%d" % value, address)
+        if wait is True:
+            pos = self._wait_FE(address)
```
YannickDieter, line 79: can we please change "FE" to something more meaningful?

basil/HL/mercury.py
```diff
@@ -56,8 +73,42 @@ def get_channel(self, address=None):
...
-    def move_relative(self, value, address=None):
+    def move_relative(self, value, precision=500, address=None, wait=True):
+        target = self.get_position(address=1) + value
         self._write_command("MR%d" % value, address)
+        if wait is True:
+            pos = self._wait_FE(address)
+            if abs(pos - target) <= precision:
+                print("At position", pos, "Traget at", target)
+            else:
+                print("Warning: Precision not reached")
```
YannickDieter, line 93: Same as above

basil/HL/mercury.py
```diff
@@ -56,8 +73,42 @@ def get_channel(self, address=None):
...
 
-    def set_position(self, value, address=None):
+    def set_position(self, value, precision=500, address=None, wait=True):
         self._write_command("MA%d" % value, address)
+        if wait is True:
+            pos = self._wait_FE(address)
+            if abs(pos - value) <= precision:
+                print("At position", pos, "Traget at", value)
+            else:
+                print("Warning: Precision not reached")
```
YannickDieter, line 83: Should be something like: "Target not reached", no? Maybe one could print the target, actual value, and prec, too.

YannickDieter (changes requested):
Looks good to me, except the minor comments. In general it would also be good to change to [str.format](https://docs.python.org/3/library/stdtypes.html#str.format).

YannickDieter:
Thanks for the update. Actually we still should replace the `prints` with proper loggers like it is done for example [here](https://github.com/SiLab-Bonn/basil/blob/9982f223891680d8d6d68315f8b651f1e3405b30/basil/HL/sram_fifo.py#L8)

basil/HL/mercury.py
```diff
@@ -56,8 +75,42 @@ def get_channel(self, address=None):
...
+        logger.info("Edge found at position: {pos}".format(pos=pos))
+
+    def _wait(self, address=None):  # waits until motor stops moving
+        logger.info("Moving! Starting position: {pos}".format(pos=self.get_position(address)))
+        done = False
+        while done is False:
+            a = self.get_position(address)
+            time.sleep(1)
+            b = self.get_position(address)
+            if a == b:
```
cbespin, line 112: Is this precise enough? I assume, some tolerance like one per mill might be good to account for numeric inaccuracies? Just wondering as this is typically the case for such that things.

YannickDieter (approved):

cbespin:
Since quite some lines are added that are not (can not?) be covered by tests, the coverage decreases, but we should ignore that.

YannickDieter:
Yeah, this was also my thought. So lets merge ;)

PR #217, opened 2024-04-24
ENH: modified get_die() for ITkPix wafermap compatibility

mjmucha:
Modified get_die() (added die number mapping) so that the SENTIO prober is compatible with ITkPix wafermap

### DIFF
--- a/basil/HL/SentioProber.py
+++ b/basil/HL/SentioProber.py
@@ -48,9 +48,19 @@ class SentioProber(HardwareLayer):
 
     def get_die(self):
         ''' Get chip index '''
-        print("Check if return order of col and row is correct and consistent with other pc drivers.")
-        values = self._intf.query("map:die:get_current_index").split(",")
-        print(int(values[-2]), int(values[-1]))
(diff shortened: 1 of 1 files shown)


cbespin (changes requested):
What is the purpose of mapping here? Shouldn't this be done on the user-side while basil concentrates on the actual device interface and returns column/row information from the machine? @YannickDieter What do you think as a second opinion?

cbespin:
Let's wait for @YannickDieter's opinion on the mapping, as I am not familiar with the probe station and the use case

mjmucha:
I moved now the different die position mappings for the ITk probing sites to a wrapper method inside [bdaq](https://gitlab.cern.ch/silab/bdaq53/-/commit/e13a71569340342d11d3d26bd42d692828500a87).

This PR can now be closed. I will open a new one for the removal of custom mappings/current die position output in basil (SignatoneProber.py and SentioProber.py)

cbespin:
This can be closed in favor of #218, right?

PR #220, opened 2024-05-07
Flowmeter

YannickDieter:


YannickDieter:
Can you please comment a bit on the changes? To me it looks like a lot is simplyfied (most of the code is removed), but it still works?

YannickDieter:
ping

matthias-schuessler:
I can't really comment on this, because this branch is entirely Antonios and Thomas work. I never worked with this branch and our current module qc environment does not use the changes in it. I can look into this.

cbespin:
@matthias-schuessler Why do you always force-push? If it does not work with a regular push, your local history is messed up.

matthias-schuessler:
@cbespin I did this, since I just performed a rebase of flowmeter onto the current version of master (basil release 3.3.0). From what I gathered from documentation, a force-push is necessary after a rebase.

matthias-schuessler:
> Yes, please have a look. In the longterm we want to use the flowmeter for our setup and it would be nice to check if the code changes actually work. Thanks! :)

The flowmeter changes in here are working properly after rebase and can be considered ready for merge.

cbespin:
You do not need a force-push after a rebase. This is the idea of a rebase ;) force-push overwrites the remote history. Always try with a normal push and if you encounter errors, deal with them first before considering a force-push.

I added myself as reviewer and will take a look as soon as possible

basil/HL/bronkhorst_elflow.py
```diff
@@ -16,160 +17,100 @@
...
+
+    def get_valve_output(self):
+        self._intf.write(self.CMDS['get_valve'])
+        ret = int(self.read()[11:], 16)
+        return ret * 61.7 / 10345949  # converts int in percentage
 
     def set_setpoint(self, value):
-        cmd = [1, 1, 0x21, (value >> 8) & 0xFF, value & 0xFF]
-        self.write(cmd)
+        """value range from 0 - 32000
```
cbespin, line 61: Is the range exactly 32000 or 32767 aka 2^15? The command seems to allow 2^16 actually? Iirc, there is a large range that does not influence the effective the setting, is it this one? Not sure if we should print a warning or info or ignore it, since the user should not care? @YannickDieter

basil/HL/bronkhorst_elflow.py
```diff
@@ -16,160 +17,100 @@
...
-        elif ret[0] == 2 and ret[1] == cmd[1] and ret[2] == cmd[2]:
-            return ((ret[3] << 8) & 0xFF00) | (ret[4] & 0xFF)
-        else:
-            logger.debug("ELFLOW.set_setpoint() ret error ret=%s" % str(ret))
-            return -1
-
-    def set_control_mode(self, value):
+        answer_in_hex = ret[11:]  # read from the 11th digits to translate what point is set
+        answer = int(answer_in_hex, 16)
+        return answer
```
cbespin, line 78: I think this might even be less "confusing" if you do `answer = int(ret[11:], 16)  # convert reply from hex to integer`

basil/HL/bronkhorst_elflow.py
```diff
@@ -16,160 +17,100 @@
...
-            return struct.unpack('!f', chr(ret[3]) + chr(ret[4]) + chr(ret[5]) + chr(ret[6]))[0]
-        else:
-            logger.debug("ELFLOW.get_valve_output() ret error ret=%s" % str(ret))
-            return -1
-
-    def get_measure(self):
-        cmd = [4, 1, 0x21, 1, 0x20]
-        self.write(cmd)
+        answer_in_hex = ret[11:]        # read from the 11th digits to translate what mode is on
+        answer = int(answer_in_hex, 16)
```
cbespin, line 96: See above

basil/HL/bronkhorst_elflow.py
```diff
@@ -16,160 +17,100 @@
...
-        if len(ret) != 5:
-            logger.debug("ELFLOW.set_setpoint() data lenth error ret=%s" % str(ret))
-            return -1
-        elif ret[0] == 2 and ret[1] == cmd[1] and ret[2] == cmd[2]:
-            return ((ret[3] << 8) & 0xFF00) | (ret[4] & 0xFF)
-        else:
-            logger.debug("ELFLOW.get_valve_output() ret error ret=%s" % str(ret))
-            return -1
+        answer_in_hex = ret[11:]  # read from the 11th digits to translate what the capacity is
+        cap_100 = struct.unpack('!f', bytes.fromhex(answer_in_hex))[0]
```
cbespin, line 107: `cap_100` is not really conclusive. Is this the maximum capacity in percent? Maybe add a comment or find a more descriptive variable name

cbespin (changes requested):
Nice PR, nice cleanup and comments. Makes the code much more readable! I have a few questions and suggestions that I am open to discuss.

Also nice to see a new example, thanks!

basil/HL/bronkhorst_elflow.py
cbespin: I heard @SilasM2001, you know more about that?

SilasM2001: I worked with the new version in the flowmeter branch. I can not commend on the previous code. The manual says setpoint has a range of 0 to 32000, with 32000 being 100%. In my experience the range works fine. The module testing setup of Matthias uses a large range (1000-25000 or so, if i recall correctly) and i did not observe a large range that does not influence the effective setting.

basil/HL/bronkhorst_elflow.py
matthias-schuessler: This is changed!

basil/HL/bronkhorst_elflow.py
SilasM2001: I just wanted to say again: The manual says setpoint has a range from 0 to 32000, 32000 being 100%. I dont know why the old command allowed values up to 32767.

basil/HL/bronkhorst_elflow.py
SilasM2001: The capacity100 command gives the maximum capacity in the selected capacity unit (This should be L/min). However one can check the unit by using the capacityunit command and by converting the responded hex code into a string. This then should give the unit.

cbespin: @SilasM2001 Just to confirm: The `cap_100` as it is used here is the percentage of the measured flow compared to the maximum available flow (in the given unit)? I.e. the maximum flow is 500 L/min and we measure 100 L/min, the function returns 20?

SilasM2001: No, the cap_100 represents the 100 % value, expressed in the Capacity Unit of the selected fluid. This value changes for different fluids, temperatures, pressures. For example cap_100 returns 100L/min, this would represent the maximum flow. By measuring the flow (this returns a value from 0 to 32000) one can convert this (since 32000 = 100% flow) to the actual flow in L/min: Actual_flow=Measured_flow/32000*cap_100

matthias-schuessler: I added a comment explaining this. Could you please check if I got it right? Then I guess this should be resolved...

cbespin: Oh boy...thank you both! I guess it should be fine

PR #225, opened 2024-05-21
Fix deprecation errors/warnings

cbespin:
Fixes _pkg_resources_ deprecation that is also relevant for importing basil and numpy casting deprecation in tests. The former effectively raises the minimum python version to 3.8, which makes sense anyway, given the support from large packages such as numpy and scipy.


cbespin:
I see there is still one warning for python3.11 and above, but the fix requires increasing the minimum python version to 3.9. Maybe we should leave it as it is for the moment and monitor the development

YannickDieter (approved):

PR #228, opened 2024-07-16
ADD: added get_source_voltage command

mjmucha:
Added the command `get_source_voltage` for the Keithley 2410 sourcemeter.


cbespin:
Not much to review or check, I guess. Does the programming manual also cover other devices? If so, would be cool to add the function to those as well. (No need to do it for all sourcemeters, only if the document at hand is valid for others anyway).

PR #229, opened 2024-08-14
Bugfixes for automatic port identification

matthias-schuessler:
Fixes a bug in findUSB.py and changes query pattern to allow it to work with HMP4040 PSU

### DIFF
--- a/basil/utils/USBBinds.py
+++ b/basil/utils/USBBinds.py
@@ -36,6 +36,7 @@ def query_identification(rm, resource, baud_rate, read_termination=None, write_t
         reply = inst.query("*IDN?", delay=0.1)
     return reply
 
+
 def find_usb_binds(rm, log,
                    instruments,
                    binds_to_skip=[],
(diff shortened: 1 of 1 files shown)


basil/utils/USBBinds.py
```diff
@@ -30,8 +30,11 @@ def query_identification(rm, resource, baud_rate, read_termination=None, write_t
     inst.baud_rate = baud_rate
     inst.read_termination = read_termination
     inst.write_termination = write_termination
-
-    return inst.query("*IDN?", delay=0.5)
+    try:
+        reply = inst.query("*IDN?", delay=0.1)
+    except pyvisa.VisaIOError:
+        reply = inst.query("*IDN?", delay=0.1)
+    return reply
```
cbespin, line 38: Why do you need a second try that is done in an `except` block? Does it always work on the second try and if not: will the whole script will crash? Or we add a loop with `x` tries that breaks when there is a reply?

cbespin:
Thank you! I will or you can merge this when the pipeline succeeded

cbespin (approved):

PR #230, opened 2024-09-24
MAINT: Prevent OverflowError in multiplication

magruber:
In newer numpy versions the multiplication of ndarrays containing uint8 results in uint8. Here this leads to an OverflowError. Therefore changing to an uint64 before the multiplication prevents this.

### DIFF
--- a/basil/utils/utils.py
+++ b/basil/utils/utils.py
@@ -41,7 +41,7 @@ def bitarray_to_byte_array(bitarr):
     ba = bitarray(bitarr, endian=bitarr.endian())
     ba.reverse()  # this flip the byte order and the bit order of each byte
     bs = np.frombuffer(ba.tobytes(), dtype=np.uint8)  # byte padding happens here, bitarray.tobytes()
-    bs = (bs.astype(np.uint64) * 0x0202020202 & 0x010884422010) % 1023
+    bs = (bs * np.uint64(0x0202020202) & 0x010884422010) % 1023
     return array('B', bs.astype(np.uint8))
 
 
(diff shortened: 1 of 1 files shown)


basil/utils/utils.py
```diff
@@ -41,7 +41,7 @@ def bitarray_to_byte_array(bitarr):
     ba = bitarray(bitarr, endian=bitarr.endian())
     ba.reverse()  # this flip the byte order and the bit order of each byte
     bs = np.frombuffer(ba.tobytes(), dtype=np.uint8)  # byte padding happens here, bitarray.tobytes()
-    bs = (bs * 0x0202020202 & 0x010884422010) % 1023
+    bs = (bs.astype(np.uint64) * 0x0202020202 & 0x010884422010) % 1023
```
cbespin, line 44: I would argue it is a bit easier to follow if you do `bs * np.uint64(0x0202020202)` to make the casting more explicit and keep `bs` an array of `uint8` aka bytes. Performance seems to be about the same with a very simple test.

cbespin (approved):
Nice catch! Only a short comment, maybe personal style and happy to discuss if you think differently

basil/utils/utils.py
magruber: Ok, I agree.

cbespin:
@magruber Could you cherry-pick this commit (66fcf32ca9d40008e01e3abca8913538fbb36c67) from https://github.com/SiLab-Bonn/basil/pull/231 into your fork and push it to update this PR? Then I will merge it and it should not cause too much trouble with [231](https://github.com/SiLab-Bonn/basil/pull/231)

PR #231, opened 2024-10-08
ENH: Fast TDL TDC

MarcoVogt:
TDC implementation with sub-ns resolution based on tapped delay lines (TDL).


cbespin:
@MarcoVogt I took some of these commits and made them their own PR, especially since #230 already existed. Maybe rebase (or merge) before you continue coding to prevent ugly merge conflicts later

tests/test_SimTdl_Tdc.py
```diff
@@ -120,53 +120,53 @@ def test_tdc(self):
...
-#
-#     def test_tdc_delay_overflow(self):
-#         pass
-#
-#     def test_tdc_delay_late_trigger(self):
-#         pass
-#
-#     def test_tdc_arm(self):
-#         pass
+    def skip_broadcasting(self):
```
cbespin, line 123: You can use the `@unittest.skip` decorator to skip tests without renaming them or commenting out. Also adds only one modified line to git ;)

MarcoVogt: Good to know. Fixed :)

cbespin:
Thanks! I will give this a review as soon as possible! There are currently no tests covering any of the `tdl_tdc`, right?

cbespin:
@MarcoVogt I am currently implementing this in a full-fledged FPGA firmware and wondering: why are the timestamps only 16bit. According to the readme, in all occurrences where they are written, they are padded with 9bits of 0s that one could use for more timestamp bits. This would help with correlating TDC words to pixel hit information.
Or am I not seeing any showstoppers to extend the timestamp to 25 bits?

PR #234, opened 2024-11-13
Refactored iseg hardware layer

mjmucha:
- Added get_* and set_* methods for compatibility reasons.
- Deprecated decorated getter/setter properties, but they are still available for backward compatibility.
- Refactored code by replacing setup_ps() with init().
- Added get_current_trip* commands for different Ampere ranges (default Ampere range is now mA).


cbespin (approved):
I don't remember the exact reason to move away from `@property`, but there was one. Rest looks good to me, thanks!

PR #235, opened 2024-11-13
Support of USB bind identification for iseq power supplies

mjmucha:
The iseq SHQ 122M does not support the standard `*IDN?` query. Therefore this PR adds the functionality to query/read the identifier of the SHQ 122M via the `#` command. Additionally included is a small fix for an bug with memorized_binds.

### DIFF
--- a/basil/utils/USBBinds.py
+++ b/basil/utils/USBBinds.py
@@ -46,7 +46,7 @@ def query_identification(rm, resource, baud_rate, read_termination=None, write_t
 def find_usb_binds(rm, log,
                    instruments,
                    binds_to_skip=[],
-                   memorized_binds=[],
+                   memorized_binds={},
                    timeout=1000 * 4
                    ):
     """
@@ -105,7 +105,7 @@ def find_usb_binds(rm, log,
             try:
                 log.debug(f"Trying {res} with baud rate {instrument['baud_rate']}")
 
-                if any(res in bind for bind in memorized_binds):
+                if any(res in bind for bind in memorized_binds.keys()):
                     log.debug(f"Found memorized bind {res}")
                     result = memorized_binds[res]
                 else:
@@ -114,7 +114,7 @@ def find_usb_binds(rm, log,
                     else:
                         result = query_identification(rm, res, instrument['baud_rate'], instrument['read_termination'], instrument['write_termination'], timeout=timeout)
 
-                    memorized_binds.append({res, result})
+                    memorized_binds.update({res: result})
 
                     log.debug(f"Found {result.strip()}")
 
--- a/basil/utils/USBBinds.py
+++ b/basil/utils/USBBinds.py
@@ -9,7 +9,7 @@ import pyvisa
 import ruamel.yaml
 
 
-def query_identification(rm, resource, baud_rate, read_termination=None, write_termination=None, timeout=1000 * 5):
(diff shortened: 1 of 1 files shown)


basil/utils/USBBinds.py
```diff
@@ -31,12 +31,16 @@ def query_identification(rm, resource, baud_rate, read_termination=None, write_t
...
         # This retries the query a second time, since some devices do not answer the first time.
         # If a second exception arrises, it will be handled in calling function.
-        reply = inst.query("*IDN?", delay=0.1)
-    return reply
+        reply = inst.query(indentifer_cmd, delay=0.1)
+        
+    if read_after_query:
+        return inst.read()
+    else:
+        return reply
```
cbespin, line 43: Why is this needed?

basil/utils/USBBinds.py
```diff
@@ -9,7 +9,7 @@
 import ruamel.yaml
 
 
-def query_identification(rm, resource, baud_rate, read_termination=None, write_termination=None, timeout=1000 * 5):
+def query_identification(rm, resource, baud_rate, read_termination=None, write_termination=None, timeout=1000 * 5, indentifer_cmd="*IDN?", read_after_query=False):
```
cbespin, line 12: Typo in `indentifier`

basil/utils/USBBinds.py
```diff
@@ -105,7 +109,10 @@ def find_usb_binds(rm, log,
                     log.debug(f"Found memorized bind {res}")
                     result = memorized_binds[res]
                 else:
-                    result = query_identification(rm, res, instrument['baud_rate'], instrument['read_termination'], instrument['write_termination'], timeout=timeout)
+                    if instrument["type"].lower() == "iseg_hv":
```
cbespin, line 112: Maybe make a global "devices_with_non_default_values" dict, so that it is easier to potentially add more devices there instead of blowing up an `if` clause?

basil/utils/USBBinds.py
```diff
@@ -105,7 +109,10 @@ def find_usb_binds(rm, log,
...
+                        result = query_identification(rm, res, instrument['baud_rate'], instrument['read_termination'], instrument['write_termination'], timeout=timeout, indentifer_cmd="#", read_after_query=True)
```
cbespin, line 113: Same typo as above

basil/utils/USBBinds.py
```diff
@@ -228,6 +235,8 @@ def modify_basil_config(conf, log, skip_binds=[], save_modified=None):
             port = None
 
         instruments.append({
+            "name": tf["name"],
+            "type": "", # Standard instruments with *IQN? command
```
cbespin: What is happening here exactly? Is this a way to define different devices with non-`*IDN?` identifier commands?

cbespin (changes requested):

basil/utils/USBBinds.py
mjmucha: Because the SHQ 122m returns as query response the command string as check if the command was executed successfully. To get the return value of an command (or in this case the identification string) one has to additionally perform a read operation.

basil/utils/USBBinds.py
mjmucha: This is done to identify serial transfer layers which uses implemented non-`*IDN?` devices. Without the addition of "name" and "type" in this temporary instrument dictionary the linking in the code block below would not work and thus also not the logic which you commented in your third comment.

**TL;DR:** The serial transfer layer does not know if it is used for e.g. an SHQ 122M. Therefore this part is needed to identify an instrument correctly.

PR #238, opened 2025-02-27
Use board number instead of address in the Mercury class

### LINKED ISSUE #237: Mercury controller board number misused as device (address) number
The board number that is read back with the `TB` command is misused as the controller address in the Mercury class. As you can see below in the screenshots from the manual of the controller `board_number = controller_address - 1`. 

Image
Image

I don't know how heavily this class is used among projects, but there are two solutions here:

1. Replace `address` with `board_number` to reflect the manual instructions. This will not break things, but the user SHOULD do the calculation in their head -> `board_number = controller_address - 1`.
2. Fix the class so it really uses the actual hardware `address` you set with the switches. However, this will break running systems.

What do you think @cbespin , @YannickDieter , @MarcoVogt ?

Giakou:
Fixes #237

### DIFF
--- a/basil/HL/mercury.py
+++ b/basil/HL/mercury.py
@@ -16,7 +16,7 @@ class Mercury(HardwareLayer):
     '''Driver for the Physiks Instruments Mercury Controller.
     A protocoll via RS 232 serial port is used with 9600/19200/38400/115200 baud rate. The baud rate
     can be set with dip switches, as well as a hardware address to distinguish several
-    devices on one line. The write termination is CR and the read ends with '\r', '\n', '\x03'.
+    devices on one line. Keep in mind that the address that is set via the DIP switches corresponds to (board number + 1). Check p.45 of the overall manual below on how to properly set the address of each controller and what the corresponding board number is. The write termination is CR and the read ends with '\r', '\n', '\x03'.
     Therefore '\x03' has to be set a read termination in the transport layer! And a read termination
     of 0.1 s should be set!
     Despite the manual telling SCPI compatibility, this is not correct for our devices with our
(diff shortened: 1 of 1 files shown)


basil/HL/mercury.py
```diff
@@ -16,7 +16,7 @@ class Mercury(HardwareLayer):
     '''Driver for the Physiks Instruments Mercury Controller.
     A protocoll via RS 232 serial port is used with 9600/19200/38400/115200 baud rate. The baud rate
     can be set with dip switches, as well as a hardware address to distinguish several
-    devices on one line. The write termination is CR and the read ends with '\r', '\n', '\x03'.
+    devices on one line. Keep in mind that the address that is set via the DIP switches corresponds to (board number + 1). Check p.45 of the overall manual below on how to properly set the address of each controller and what the corresponding board number is. The write termination is CR and the read ends with '\r', '\n', '\x03'.
```
cbespin, line 19: We do not enforce line break rules, but can you change this to reflect the line length of the others, please?

basil/HL/mercury.py
```diff
@@ -28,11 +28,11 @@ def __init__(self, intf, conf):
...
-        self._addresses = []
-        for a in range(16):  # Check all possible addresses
-            self.write(bytearray.fromhex("01%d" % (a + 30)) + "TB".encode())  # Tell board address
-            if self.get_address(a):
-                self._addresses.append(a)
+        self._board_numbers = []
+        for b in range(16):  # Check all possible board numbers
+            self.write(bytearray.fromhex("01%d" % (b + 30)) + "TB".encode())  # Tell board number
+            if self.get_board_number(b):
+                self._board_numbers.append(b)
```
cbespin, line 35: Personally, I refrain from using single-character variables; maybe `board_n` or even `board_number` is a bit easier to read. Just wanted to point it out, this is no actual change request. Hit "resolve" if you want to keep it as it is.

cbespin (approved):
Thanks, looks good to me!

PR #239, opened 2025-03-26
Add requirement introduced in #233

cbespin:
`GitPython` is required since #233 but has not been added to `requirements.txt`.


cbespin:
@YannickDieter @MarcoVogt I spotted that mistake since up to now we were just lucky that the projects running pipelines with basil required `GitPython` on their own. I can not review my own PR, so I need someone to "check" it please

MarcoVogt (approved):

PR #241, opened 2025-08-18
Adding firmware for Grove 4-channel relay

matthias-schuessler:
Adding Arduino firmware to control Grove 4-channel SPDT relay using an Arduino Mega. This is based on the Arduino relay board and uses the same commands.

### DIFF
--- a/basil/HL/arduino_grove_4channel_relay.py
+++ b/basil/HL/arduino_grove_4channel_relay.py
@@ -43,12 +43,12 @@ class RelayBoard(ArduinoBase):
 
         ret = self.query(self.create_command(self.CMDS['write'], channel, value))
 
-        error = False
-
         if channel == 99 and int(ret) != value * 1111:
             error = True
         elif channel != 99 and format(int(ret, 2), '04b')[-channel] != str(value):
             error = True
+        else:
+            error = False
 
         if error:
             raise RuntimeError('Got no or wrong response from Arduino!')


cbespin:
Since I did the last commit, and Github will not let me easily, could one of @YannickDieter @rpartzsch take a look?

basil/HL/arduino_grove_4channel_relay.py
```diff
@@ -0,0 +1,57 @@
...
+
+    CMDS = {
+        'read': 'R',
+        'write': 'W',
+        'delay': 'D'
+    }
+
+    ERRORS = {
+        'error': "Serial transmission error"  # Custom return code for unsuccesful serial communciation
+    }
```
cbespin: I do no see this being used.

basil/HL/arduino_grove_4channel_relay.py
```diff
@@ -0,0 +1,57 @@
...
+
+from basil.HL.arduino_base import ArduinoBase
+
+
+class RelayBoard(ArduinoBase):
+    '''
+    Implement functions to control the Arduino digital IO using the Basil Arduino firmware.
+    '''
+
+    CMDS = {
```
cbespin: Standard-style would be to use lowercase variable names except for global variables (outside of any scope)

cbespin (changes requested):

basil/HL/arduino_grove_4channel_relay.py
matthias-schuessler: self.ERRORS is used in the ArduinoBase Class (super or this)

basil/HL/arduino_grove_4channel_relay.py
matthias-schuessler: These are class wide dictionaries, with the naming coming from ArduinoBase

PR #250, opened 2026-01-09
Fixes for Cocotb v2

MarcoVogt:
Several small fixes to make Basil compatible with Cocotb version 2.

Migration guide: https://docs.cocotb.org/en/v2.0.1/upgrade-2.0.html

### DIFF
--- a/examples/MIO/src/example.v
+++ b/examples/MIO/src/example.v
@@ -26,8 +26,6 @@ module example (
     output wire LED4,
     output wire LED5,
 
-    inout wire FPGA_BUTTON,
-
     inout wire SDA,
     inout wire SCL
 );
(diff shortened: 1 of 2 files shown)


examples/MIO/src/example.v
```diff
@@ -87,8 +87,6 @@ assign SDA = 1'bz;
 assign SCL = 1'bz;
 assign DEBUG_D = 16'ha5a5;
 
-`ifdef COCOTB_SIM
-    assign FPGA_BUTTON = 0;
-`endif
+assign FPGA_BUTTON = 1'b0;
```
cbespin: Is this always fine, also when synthesizing the example?

PR #253, opened 2026-02-18
Reduce number of bits for address pointer

cbespin:
In Vivado 2025.2, I am still getting errors regarding the pointer size. When inferring BRAM, the instantiated RAMB36E1 block has a maximum address size of 16bit, so I adjusted it in the `generic_fifo` accordingly.
While it should not matter for distributed RAM, the address space is still large enough for any reasonable use case at the moment, I think.


MarcoVogt (approved):
Looks good

PR #269, opened 2026-05-08
Fix port width of wrapper in generic_fifo

cbespin:
The port width was incorrectly 1 for all signals in the wrapper, affecting compilation with old designs that still use the `gerneric_fifo` module name.


rpartzsch (approved):
Nice

PR #273, opened 2026-05-27
Include missing firmware READMEs

### LINKED ISSUE #267: Update Modules in Documentation
We should update the firmware modules in the documentation. I think we are missing many here. This should be quite easy by linking the README in [modules](https://github.com/SiLab-Bonn/basil/blob/master/docs/modules.rst).

kcaisley:
Closes #267

Adding the missing firmware README files in docs/modules.rst and renames the mixed-case Readme.rst filenames to README.rst.


rpartzsch (changes requested):
Nice, I think the `bram_fifo` is still missing. But this should be all of them.

kcaisley:
Ah okay, so I think this needs to be cleared up:

The BRAM docs currently look like the below snippet, but that's mixing up the BRAM and SRAM fifos, right?

```rst
.. !BRAM_FIFO!

.. include:: ../basil/firmware/modules/bram_fifo/README.rst

**Driver:**

.. automodule:: basil.HL.sram_fifo

.. autoclass:: sram_fifo
    :members:
```

kcaisley:
I think I've cleaned this up now, and added missing doc strings for the drivers of the `sram_fifo` and `bram_fifo`.

rpartzsch:
You are right. Looks good to me.

rpartzsch (approved):

PR #274, opened 2026-05-27
WIP: Minor simulation missing

padeken:
Hi I would like to have these sim stand-ins in basil. I think they are useful for more projects.

### DIFF
--- a/basil/firmware/modules/utils/PLLE2_sim.v
+++ b/basil/firmware/modules/utils/PLLE2_sim.v
@@ -1,0 +1,21 @@
+`ifdef COCOTB_SIM
+module PLLE2_sim (
+    input  wire CLKIN1,
+    output wire CLKOUT0,
+    output wire CLKOUT1,
+    output wire CLKOUT2,
+    output wire CLKOUT3,
+    output wire CLKOUT4,
+    output wire CLKOUT5,
+    output wire LOCKED
+);
+    // Simple behavioral model: forward input clock to all outputs and indicate locked
+    assign CLKOUT0 = CLKIN1;
+    assign CLKOUT1 = CLKIN1;
+    assign CLKOUT2 = CLKIN1;
+    assign CLKOUT3 = CLKIN1;
+    assign CLKOUT4 = CLKIN1;
+    assign CLKOUT5 = CLKIN1;
+    assign LOCKED = 1'b1;
+endmodule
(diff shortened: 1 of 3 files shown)


cbespin:
Funnily enough, part of it (at least the PLL one) is included in #272 that is about to be merged the next days. I would wait until this one is done and then check again. The RAM block will probably still be needed as I don't think it is included in #272, but I would have to check (it is quite a big one..)

kcaisley:
Actually, after chatting with @awalsemann I actually removed the PLL block. Since my version was also just a black box, we couldn't think of a good reason to keep it. But I'll chat with @padeken, since if he wants it there's obviously a use case!

padeken:
Hi Ned,
yes then we move it to your merge. I just need a dummy for simulation and I just used the input as output, because it worked for my simplified simulation.

cbespin:
Well, if you need some sort of PLL in a simulation with more than one clock output, maybe we can make a more flexible simulation solution with instantiating clock dividers in the PLL replacement module?

padeken:
Yes, and writing this I can also provide the not sim module for the RAM, which should be more efificent, but I need to test this first

cbespin:
So you need the RAM simulation module to work on a more advanced PLL one? Then we could split this and merge the RAM first.
Do you instantiate the RAM module at some point or do you use an inferred version? In this case, one would not need the simulation module, no? Just wondering, because I did not run into this issue yet with designs that infer the primitive, but that might be on my end.

padeken:
I think I will talk to Ned and we will make a better PLL sim and the RAM maybe I will test first and then put it here.

awalsemann:
I belive if you need a more advanced model for a PLL, just use an encrypted model from AMD or use one of the exitsing open models available, like this one:  https://github.com/nmi-leipzig/sim-x-pll 
I will not stop you from developing such models on your own, but I don't see the point in us reinventing the wheel.

cbespin:
In light of the [behavioral modules](https://github.com/SiLab-Bonn/basil/tree/behavioral-modules) branch, is this PR still relevant or can it be closed? I see modules that @awalsemann mentioned in there @kcaisley . @padeken: Depending on your urgency, maybe we can move this to their own PR or get going to merge it.

padeken:
This fixed it. Thanks!

PR #277, opened 2026-05-28
Fix redundant dependencies that might break builds

cbespin:
#272 introduced duplicate dependencies in the `pyproject.toml`, that should be optional ones in any case. This might break builds, if the environment is not set up and targeting any cocotb business. @kcaisley

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,10 +27,6 @@ dependencies = [
     "GitPython",
     "six",
     "ruamel.yaml",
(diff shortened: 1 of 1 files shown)


rpartzsch (approved):

kcaisley:
> #272 introduced duplicate dependencies in the `pyproject.toml`, that should be optional ones in any case. This might break builds, if the environment is not set up and targeting any cocotb business. @kcaisley

Okay, thanks for pointing this out!

PR #280, opened 2026-07-13
Fix spi bram

kcaisley:
In pull request #272, I replaced the explicit outdated instance of `RAMB16_S1_S9` with an attempt to use the `ram_style` attribute, allowing a behavioral description during simulation that would then map to a RAMB18E1 during Vivado synthesis.

Turns out that I did this wrong, and that ram_style = "block" alone cannot make an arbitrary memory description infer BRAM. Since the usage of the RAM here is a little fancy, with different widths for the two ports, it was actually just mapping to LUT RAM.

I have now read the [Vivado UG901 document](https://docs.amd.com/r/en-US/2026.1/ug901-vivado-synthesis/Dual-Port-Asymmetric-RAM-When-Read-is-Wider-than-Write-Verilog), though, which explains how to properly use this attribute together with a supported asymmetric RAM inference template. 

This fix changes the behavioral description to follow that template. The memory is described as a 16,384 x 1-bit array. The narrow port accesses individual bits directly, while the wide port uses a fixed loop to read or write eight consecutive bits as one byte. Verilator and Icarus can simulate it behaviorally, while Vivado recognizes the same description as block RAM (each of the two instances in `spi_core` maps one `RAMB18E1`).

In the synthesis report, one now sees the following:

```
Block RAM: Final Mapping Report
+-----------------------+------------+------------------------+---+---+------------------------+---+---+------------------+--------+--------+
|Module Name            | RTL Object | PORT A (Depth x Width) | W | R | PORT B (Depth x Width) | W | R | Ports driving FF | RAMB18 | RAMB36 | 
+-----------------------+------------+------------------------+---+---+------------------------+---+---+------------------+--------+--------+
|blk_mem_gen_8_to_1_2k: | ram_reg    | 16 K x 1(READ_FIRST)   | W | R | 2 K x 8(READ_FIRST)    | W | R | Port A and B     | 1      | 0      | 
|blk_mem_gen_8_to_1_2k: | ram_reg    | 16 K x 1(READ_FIRST)   | W | R | 2 K x 8(READ_FIRST)    | W | R | Port A and B     | 1      | 0      | 
+-----------------------+------------+------------------------+---+---+------------------------+---+---+------------------+--------+--------+
```

### DIFF
--- a/basil/firmware/modules/spi/blk_mem_gen_8_to_1_2k.v
+++ b/basil/firmware/modules/spi/blk_mem_gen_8_to_1_2k.v
@@ -4,19 +4,27 @@
  * SiLab, Institute of Physics, University of Bonn
  * ------------------------------------------------------------
  *
- * Asymmetric true dual-port BRAM: 2048 x 8-bit (Port A) / 16384 x 1-bit (Port B).
- * Total capacity: 16,384 bits → inferred as RAMB18E1 on 7-series (Kintex-7).
+ * Asymmetric dual-port memory: 2048 x 8-bit (Port A) / 16384 x 1-bit (Port B).
+ * Total capacity: 16,384 bits → intended to infer one 7-series RAMB18E1 in
+ * block-RAM memory mode. See AMD/Xilinx UG473, 7 Series FPGAs Memory Resources:
+ * https://docs.amd.com/v/u/en-US/ug473_7Series_Memory_Resources
  *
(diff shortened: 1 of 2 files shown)


cbespin (approved):
Since it is included and instantiated in `spi`, and the tests pass, I assume there is no download incompatibility. I am not knowledgeable enough to judge the code, but it looks good.

PR #282, opened 2026-08-19
Ping available I2C address

rpartzsch:
Pings all I2C address and checks for an acknowledged bit.
This is useful to check which I2C devices are connected and respond to basic pings.


cbespin (approved):
Very useful, thanks!

PR #283, opened 2026-08-19
Banish the ghost of codecov

kcaisley:
I looked into these randomly appearing Codecov bot messages, and it turns out the integration has actually been active in this repository since 2020 ([it was added in this commit](https://github.com/SiLab-Bonn/basil/commit/389f4f74d4f1ddd271301660c80ae02807fea844)). It has occasionally worked since then, for example, it commented on [PR #251](https://github.com/SiLab-Bonn/basil/pull/251#issuecomment-3744820848), but most pull requests never receive a report.

The integration is configured in `.github/workflows/regression-tests.yml`. Because we do not have a `CODECOV_TOKEN` configured, it uses Codecov's legacy tokenless upload method. According to [the docs](https://docs.codecov.com/docs/codecov-tokens#legacy-upload-methods): "This method shares a global rate limit with other users" and is documented as unreliable, which explains why the integration only works intermittently.

So yeah... rather than re-setting up and maintaining another external account and access token for an unreliable service, this replaces Codecov with a simple coverage report produced directly by GitHub Actions.

Thoughts?

### DIFF
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -22,7 +22,7 @@ jobs:
         python-version: "3.12"
     - uses: pre-commit/action@v3.0.1
       env:
-        SKIP: diff-coverage # Run all normal formatting hooks in CI, but skip the local diff-coverage hook because the coverage job handles it separately.
+        SKIP: coverage # Run all normal formatting hooks in CI, but skip the local coverage hook because the coverage job handles it separately.
 
   tests:
 
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -42,6 +42,7 @@ jobs:
 
           - sim: icarus
             python-version: '3.13'
+            coverage: true
 
           - sim: verilator
             sim-version: 5.044
@@ -71,7 +72,13 @@ jobs:
     - name: Install Python dependencies
       shell: bash -l {0}
       run: |
-        pip install pyvisa pyvisa-sim pytest coverage coveralls pytest-cov cocotb>=2 cocotb-bus
+        pip install pyvisa pyvisa-sim pytest cocotb>=2 cocotb-bus
+
+    - name: Install coverage dependency
+      if: matrix.coverage
+      shell: bash -l {0}
+      run: |
+        pip install pytest-cov
 
     - name: Install basil
       shell: bash -l {0}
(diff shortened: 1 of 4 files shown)


kcaisley:
Oh also, for this to work, we may need to adjust the repo or organization settings to allow github's coverage action to run: [see here](https://docs.github.com/en/code-security/how-tos/maintain-quality-code/set-up-code-coverage)

cbespin:
Enabling code quality comes with a 10€ per month additional cost. I think there are other ways to get coverage report in the repo (but maybe not on each PR). Nevertheless, I will merge this to get rid of codecov first and we can see if we find a nice way how to do it.

cbespin (approved):

cbespin:
Approved in Draft state, sorry. Let me know if and when it is ready.

kcaisley:
Okay, so I did a bit of searching, and it looks like the already installed [`coverage`](https://coverage.readthedocs.io/en/7.15.4/#) python package we have is already producing for us a coverage.xml file. On the cli that would already produce for us an output like this:

```
$ coverage report -m
Name                      Stmts   Miss  Cover   Missing
-------------------------------------------------------
my_program.py                20      4    80%   33-35, 39
my_other_module.py           56      6    89%   17-23
-------------------------------------------------------
TOTAL                        76     10    87%
```

From here, if we want to see the _change_ in coverage, relative to another commit or branch, it seems the platform/language agnostic [`diff-cover`](https://github.com/Bachmann1234/diff_cover) package can simply be run:

```
diff-cover coverage.xml --compare-branch=origin/master
```


That's a bit wordy to run though, so to see it easily locally I've added a `pre-commit` hook, so you can simply do `pre-commit run coverage` to see if your uncommitted changes have reduced coverage.

So, for example: If a new function was added in `example.py` but doesn't have tests, it would show this:
 
 ```
Coverage................................................................Passed
- hook id: coverage
- duration: 178.2s

........................................................................ [100%]
104 passed, 6 skipped, 105 warnings in 177.5s

-------------
Diff Coverage
Diff: origin/master...HEAD, staged and unstaged changes
-------------
basil/HL/example.py (25.0%): Missing lines 84-86
-------------
Total: 25%
```

Here 25% means is the percentage of lines of code which are covered by tests.
 
And to see it in github PRs, I've added an action which displays it as a non-blocking check. This last part I'm not exactly sure if I've done right, but we'll know when we try to merge if it looks right.

kcaisley:
Ah, btw the best part is, this is entirely platform portable, so it would be able to be trigged by any CI, since it's just based on `coverage` + `diff-cover` + `pre-commit`.

kcaisley:
You can see here it creates this 'neutral check' 
image

rpartzsch (approved):

PR #284, opened 2026-09-15
Update bdaq53 names to bdaq core in examples

kcaisley:
Let's update those file names to get the main tests passing again


cbespin (approved):
Thanks!

PR #285, opened 2026-09-15
Implement StdRegister readback

kcaisley:
On my `behavioral-modules` branch, I've had this fix for a while, but actually Stella needs it too now. Can we merge this? Since my other branch is too big to review quickly


kcaisley:
Hmm, I'm now uncertain if this is a good idea. Do we know the historical reason why this was not implemented? Back in [2019 Jens changed it:](https://github.com/SiLab-Bonn/basil/commit/3396223abede51cd270f351a8c275d7631522979#diff-c1aafe3c02b5a744c06915b29ef5567a15caa7955e35ec58bca9513ed5e41a80L123-R123) 

I guess one thing to think about here, is that it will update the `StdRegister` object which might not be desirable? Just directly calling `get_data` would avoid this. But I'm not sure if that is a bug or feature.

kcaisley:
To be more clear, `.get_data()` fetches raw bytes from the hardware; but this `StdRegister.read()` additionally copies those bytes into the named fields. If we request to read just one named field, then only that one will be refreshed in `StdRegister`. So what I mean is:

```python
daq["gpio1"].read()  # Refresh every field
ready = daq["gpio1"]["COMP_IDELAY_RDY"]  # Access and refresh only one field
```

cbespin:
It seems reasonable to me and well-worded that `.read()` reads all hardware values into the dictionary fields. Especially, if it is needed and useful, you have my blessings :grinning:

cbespin (approved):
Feel free to merge when ready

rpartzsch (approved):

kcaisley:
Sweet okay, let's try it!

PR #286, opened 2026-09-15
Fix last remaining old-style RAM type description

cbespin:
The i2c firmware module still had the old-style type definition for BRAM. This is fixed now and consistent with all other modules explicitly requesting BRAM.


kcaisley:
Hey, so this looks good to me. I don't know that much about this `i2c` module, but one thing I think we could also fix there though is the both before and after this change, the default `parameter MEM_BYTES=1` appears to be a invalid value?

```
WARNING: [Synth 8-4767] Trying to implement RAM 'mem_reg' in registers. Block RAM or DRAM implementation is not possible;
see log for reasons.
Reason is one or more of the following :
    1: Only 1 word in RAM
WARNING: [Synth 8-4557] Cannot dissolve ram 'mem_reg' to registers because of more than 1 writes
ERROR: [Synth 8-2913] Unsupported Dual Port Block-RAM template for mem_reg
```

I guess we haven't noticed this before, because across `frida1-daq`, `obelix1-daq`, and `tj-monopix2-daq` we always are setting it to `.MEM_BYTES(32)`.

So maybe we should just set a line like this, inside `i2c.v` and `i2c_cores.v`?

```verilog
// Buffer capacity in bytes; use at least 2
// Vivado cannot infer with MEM_BYTES = 1
parameter MEM_BYTES = 32,
```

In the future, we could also take a look at [`ramb_8_to_n.v`](https://github.com/SiLab-Bonn/basil/blob/d18745fae16c6f8f5948d8c7f77efdc6b054d14d/basil/firmware/modules/utils/ramb_8_to_n.v#L53), which is a synchronous dual-port memory used by the sequencer, recorder, and JTAG modules, and [`gpac_adc_rx_core.v`](https://github.com/SiLab-Bonn/basil/blob/d18745fae16c6f8f5948d8c7f77efdc6b054d14d/basil/firmware/modules/gpac_adc_rx/gpac_adc_rx_core.v#L199-L213) which i guess is some 256 x 14-bit ADC delay buffer with synchronous reads and writes. But the latter at least I think it's fairly un-unused since it's on the gpac card.

cbespin:
> I guess we haven't noticed this before, because across frida1-daq, obelix1-daq, and tj-monopix2-daq we always are setting it to `.MEM_BYTES(32)`.

I do not know why, actually. ~Maybe 8 is a better value to infer less resources, if needed?~ e: nevermind, the smallest RAMB primitive seems to be [18 Kb in AMD 7-series](https://docs.amd.com/r/en-US/ug953-vivado-7series-libraries/RAMB18E1) so 32 is still perfectly fine. Let's go with 32, or do you have objections?

kcaisley:
Yep, I think a default of 32 would be ideal, since all projects I can see are already using that.

basil/firmware/modules/i2c/i2c_core.v
```diff
@@ -115,7 +115,7 @@ wire [ABUSWIDTH-1:0] BUS_MEM_ADD;
 assign BUS_MEM_EN = (BUS_WR | BUS_RD) & BUS_ADD >= 8;
 assign BUS_MEM_ADD = BUS_ADD-8;
 
-(* RAM_STYLE="{BLOCK_POWER2}" *)
+(* ram_style = "block" *)
 reg [7:0] mem [MEM_BYTES-1:0];
```
kcaisley: when you change the default for this, make sure to get it in the wrapper and the core block!

cbespin: There is no other occurrence of the `RAM_STYLE` attribute anywhere related to I2C, or where do you mean?

kcaisley: I mean the parameter has a default in both the `i2c_core.v` and the `i2c.v`:

```verilog
module i2c_core #(
    parameter ABUSWIDTH = 16,
    parameter MEM_BYTES = 1,
    parameter IGNORE_ACK = 0
) (
    input wire BUS_CLK,
    input wire BUS_RST,
    input wire [ABUSWIDTH-1:0] BUS_ADD,
    input wire [7:0] BUS_DATA_IN,
    input wire BUS_RD,
    input wire BUS_WR,
    output reg [7:0] BUS_DATA_OUT,

    input wire I2C_CLK,
    inout wire I2C_SDA,
    inout wire I2C_SCL

);
```

```verilog
module i2c #(
    parameter BASEADDR = 16'h0000,
    parameter HIGHADDR = 16'h0000,
    parameter ABUSWIDTH = 16,
    parameter MEM_BYTES = 1,
    parameter IGNORE_ACK = 0
) (
    input wire                 BUS_CLK,
    input wire                 BUS_RST,
    input wire [ABUSWIDTH-1:0] BUS_ADD,
    inout wire [7:0]           BUS_DATA,
    input wire                 BUS_RD,
    input wire                 BUS_WR,

    input wire I2C_CLK,
    inout wire I2C_SDA,
    inout wire I2C_SCL
);
```

kcaisley: Ah but I see the new commit, and you go both! Looks good to go for me! :)

cbespin: I thought you were referring to the `RAM_STYLE`. Adjusted both in the core and wrapper now!

kcaisley (commented):
Looks good!

kcaisley (approved):

PR #287, opened 2026-09-16
Use install extra for CI tests

### LINKED ISSUE #278: Test dependencies in workflow
The test dependencies (cocotb, coverage, ...) are installed in the workflow file, but should be installed from the pip target `testing` instead to prevent wrong versions or interference.

cbespin:
Closes #278 . The `pyproject.toml` provides an install extra for tests, which was ignored in the CI so far.


cbespin:
Updated the versions of the used Github actions as well, since some used deprecated Node.js 20

rpartzsch (approved):
Cool

